PR #2, opened 2019-11-15
Update deploy.sh

chicoxyzzy:


mathiasbynens:
I’m not sure if we want to continue publishing the website, given that most people seem to be linking to the `*.md` files anyhow. But the changes look good, so let’s merge them until we have that discussion.

PR #3, opened 2019-11-15
Update README.md

jridgewell:


README.md
```diff
@@ -2,5 +2,8 @@
 
 Published at: 
 
-- http://tc39.github.io/tc39-notes
+- http://tc39.github.io/notes
```
mathiasbynens, line 5: We should probably remove this link. Folks have been linking to the `*.md` files instead of the gh-pages hosted site (which hasn’t yet been set up for this new repo) anyhow.

PR #20, opened 2019-12-18
Add myself to delegates (from Sujitech)

Jack-Works:


Jack-Works:
@codehag hello

codehag:
Hi Jack, since we have so many participants, the convention for new members is to use a three letter acronym. The pattern is first letter of first name, first letter of last name, last letter of last name. If for example you use "Jack Works", it would be "JWS". We usually do not use pseudonyms however, so it would be great to have your english working name here.

codehag:
At the moment, a lot of people are on holidays -- it may take some time for this to be merged. There should be more activity starting January 2nd.

codehag:
Hi, it looks like there is still an issue. Jason Williams has the acronym (JWS), so you will need to use something else. 

You can see the checks here: https://github.com/tc39/notes/blob/master/scripts/check-delegates.js

Jack-Works:
I have changed to JWK

PR #47, opened 2020-05-13
Improve some links for March 2020 meeting notes

chicoxyzzy:
I've added missing links and crosslinking (between different days) for some proposals


ryzokuken (approved):
Thanks!

PR #50, opened 2020-06-02
Adding Younies Mahmoud (YMD)

rricard:


mpcsh:
@younies we went with `YMD` by default. are you okay with this or would you like to use a different abbreviation?

rricard:
@younies can you check if you want to appear as YMD in the notes. Thanks!

sffc:
YMD is also what we used in February and what we use in ECMA-402.

younies (approved):

younies:
thanks :)

PR #55, opened 2020-06-23
Change LBR to LEO (Leo Balter)

leobalter:
LBR does not reflect my initials and it causes confusion to me as it caused confusion to other people.

I don't love changing such a long history but I'd like to end this little chaos.


sffc:
Sounds good to me.  Can you also change it in the ECMA-402 repo?

robpalme (approved):

PR #58, opened 2020-07-03
feat: add Yorkie Liu(YKL) from Alibaba

pusongyang:
feat: add Yorkie Liu(YL) from Alibaba


gesa:
Hey @pusongyang acronyms should be three letters long. The only case where they are two letters long is for people who have been participating in the committee since before this policy was implemented years ago.

pusongyang:
> Hey @pusongyang acronyms should be three letters long. The only case where they are two letters long is for people who have been participating in the committee since before this policy was implemented years ago.

OK，I switch to YKL.

gesa (approved):

PR #61, opened 2020-07-21
Fix link to temporal update slides

Alexendoo:


ptomato (approved):
This seems to be correct, I believe @pipobscure rearranged things :smile:

Alexendoo:
Added two other links as well

PR #65, opened 2020-08-07
Add notes for July 2020

robpalme:
These notes will be merged tomorrow (2020-08-08). Please raise any final review comments today!

### DIFF
--- a/meetings/2020-07/july-20.md
+++ b/meetings/2020-07/july-20.md
@@ -432,7 +432,7 @@ RBN: My question would be, how would you imagine that `var` would work in that c
 SYG: Yes
 
 RBN: I’ll be honest, if I could’ve written this as `static(){`, then I would have, but that's already legal JS 
-Because it has special semantics around how `this` works, because otherwise who knows what `this` would be referring to, and then we have the complexity around what happens with return, await, yield. I saw on IRC that there’s some discussion about the class access constructions, that one, I may eventually bring back, but that one had some strange things around evaluation, since its main motivation was around private static and subclassing.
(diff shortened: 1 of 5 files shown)


ljharb (approved):

leobalter (approved):

sffc (approved):

meetings/2020-07/july-20.md
```diff
@@ -0,0 +1,543 @@
...
+
+RBN: My question would be, how would you imagine that `var` would work in that case? Would it lift out of static block into the surrounding function scope?
+
+SYG: Yes
+
+RBN: I’ll be honest, if I could’ve written this as `static(){`, then I would have, but that's already legal JS 
+Because it has special semantics around how `this` works, because otherwise who knows what `this` would be referring to, and then we have the complexity around what happens with return, await, yield. I saw on IRC that there’s some discussion about the class access constructions, that one, I may eventually bring back, but that one had some strange things around evaluation, since its main motivation was around private static and subclassing.
```
rbuckton:
```suggestion
Because it has special semantics around how `this` works, because otherwise who knows what `this` would be referring to, and then we have the complexity around what happens with return, await, yield. I saw on IRC that there’s some discussion about the class access expressions, that one, I may eventually bring back, but that one had some strange things around evaluation, since its main motivation was around private static and subclassing.
```

meetings/2020-07/july-20.md
```diff
@@ -0,0 +1,543 @@
...
+So again the main reason that it’s essentially treated kind of like a function but I can’t put parens in because it would conflict with existing syntax, and then vars ??? be hoisted out.
```
rbuckton:
```suggestion
So again the main reason that it’s essentially treated kind of like a function but I can’t put parens in because it would conflict with existing syntax, and as such vars would not be hoisted out.
```

rbuckton (changes requested):

PR #90, opened 2020-12-07
Add notes for Nov 2020

robpalme:
These notes will be merged tomorrow. Please incorporate any final review changes today by reviewing this PR.

Please note, this is the first time we are publishing notes that were largely transcribed by @bakkot's bot using Google Speech API.  As a result, they are far more verbose than previous notes, and are likely to still contain a fair amount of errors.  Please review any topics that you led.

As always, thank you also to our human note-takers/editors @rricard @bakkot @michaelficarra @codehag and any one else who contributed.

refs: https://github.com/tc39/Reflector/issues/341


michaelficarra:
@robpalme You may want to check the commit author email. It doesn't appear to be associated with your GitHub account.

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
+
+DRR: Sounds good. All right. Thanks Kevin. Thank you all and I'm done. Thank you.
+#### Conclusion/Resolution
+* Add tests.
+## __proto__ normative optional options
+Presenter: Jordan Harband (JHD)
+
+- [pr](https://github.com/tc39/ecma262/pull/2125#issuecomment-698502404)
+
+JHD:  So this is talking about PR #2125 on the spec. The current status of the pull request is that it is a normative request that moves __proto__ out of Annex B. There was an open question on it about whether it's allowed for implementations to do all to do a partial implementation of Annex B in other words, can you have __defineGetter__ and omit __defineSetter__? The editor group decided that we would land 2125 once it had had  appropriate updates without addressing the All or Nothing question, and Gus decided to add this to discuss in plenary about that question, which can be done in a separate PR or this one.
```
ljharb:
```suggestion
JHD: So this is talking about PR #2125 on the spec. The current status of the pull request is that it is a normative request that moves __proto__ out of Annex B. There was an open question on it about whether it's allowed for implementations to do all to do a partial implementation of Annex B in other words, can you have __defineGetter__ and omit __defineSetter__? The editor group decided that we would land 2125 once it had had  appropriate updates without addressing the All or Nothing question, and Gus decided to add this to discuss in plenary about that question, which can be done in a separate PR or this one.
```

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
+
+JHD: So then to modify the proposal then Mark you're suggesting that all of `__defineGetter__`, `__defineSetter__`, `__lookupGetter__`, and `__lookupSetter__` be treated as one bundle.
+
+MM: exactly.
+
+KG: But not the __proto__ accessor; the __proto__ accessor would be a separate part.
+
+MM: exactly. And in fact, I would be okay with the __defineGetter__ bundle becoming mandatory. Salesforce had a security concern that surprised me, but that is real, that makes me agree that the __proto__ accessor should remain optional. And in any case it's quite distinct from the defineGetter bundle conceptually.
+
+JHD:  So I'm going to then assume that we have consensus on making those four items one bundle and leaving the Proto accessor separate. And in the future as these questions come up we will bring them to plenary with the understanding that we are likely to decide on various little bundles ad hoc. Thanks everybody.
```
ljharb:
```suggestion
JHD: So I'm going to then assume that we have consensus on making those four items one bundle and leaving the `__proto__` accessor separate. And in the future as these questions come up we will bring them to plenary with the understanding that we are likely to decide on various little bundles ad hoc. Thanks everybody. 
```

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
+KG: But not the __proto__ accessor; the __proto__ accessor would be a separate part.
+
+MM: exactly. And in fact, I would be okay with the __defineGetter__ bundle becoming mandatory. Salesforce had a security concern that surprised me, but that is real, that makes me agree that the __proto__ accessor should remain optional. And in any case it's quite distinct from the defineGetter bundle conceptually.
+
+JHD:  So I'm going to then assume that we have consensus on making those four items one bundle and leaving the Proto accessor separate. And in the future as these questions come up we will bring them to plenary with the understanding that we are likely to decide on various little bundles ad hoc. Thanks everybody. 
+#### Conclusion/Resolution
+`__defineGetter__`, `__defineSetter__`, `__lookupGetter__`, `__lookupSetter__`, will be treated as one “bundle”: an implementation must implement either all or none of them
+`__proto__` accessor will remain independently normative optional
+Future questions of this nature will be addressed in plenary in an ad-hoc nature as items are lifted out of Annex B
+## Re-resolve unresolvable bindings in PutValue
```
ljharb, line 291:
```suggestion

## Re-resolve unresolvable bindings in PutValue
```

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
+YSV: it's cleared up in the chat.
+
+BSH: I'm just curious - it was stated that we know that if we fix this it breaks various important websites; do we know why? Do we know what is it they're doing that somehow the program is actually working even though they're setting things that didn't actually get set and yet that's okay.
+
+KG: I don't believe we know why, unfortunately.
+
+BSH: Okay. Just wondering, yeah.
+#### Conclusion/Resolution
+* Accepted, reluctantly.
+## clarifying conclusion to __proto__ normative optional agenda item
```
ljharb, line 368:
```suggestion

## clarifying conclusion to __proto__ normative optional agenda item
```

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
+
+MM: I feel strongly they should be bundled together. The way to remove the setting behavior to remove the accessor property.
+
+JHD: Does anyone else have thoughts on that? if not, I think we'll just go with bundling the two of them together.
+
+MF: My preference aligns with Mark.
+
+#### Conclusion/Resolution
+`__proto__` getter and setter will be bundled
+## Give %TypedArray% methods explicit algorithms
```
ljharb, line 380:
```suggestion

## Give %TypedArray% methods explicit algorithms
```

in general, we should make sure there's a blank line before any `#`-marked headings.

meetings/2020-11/nov-17.md
```diff
@@ -0,0 +1,729 @@
...
+
+GCL: It's cool to hear. I'd be interested to know how that affects iterator helpers,. Good to hear. 
+
+DE: [unintelligible]
+
+JHD: It's always really frustrating when the use of shiny convenient syntax brings with it a whole new vulnerability to prototypes being modified. Like for example node has I mean and this is node so it can it could always find other solutions but node has this pattern internally called primordial is where it like pulls in a fresh copy, you know pre-prepared frozen copy of prototype methods and but it's use of spreads syntax is always vulnerable to user code breaking node itself and it's really painful to try and rewrite things to not use things like spread syntax to avoid this problem. And you know, like I said, node has other alternatives because it's an engine but the same sort of thing happens in user code and like babel output and so on. It's a lot that has to be worked around. So this is a wonderful change.
```
ljharb:
```suggestion
JHD: It's always really frustrating when the use of shiny convenient syntax brings with it a whole new vulnerability to prototypes being modified. Like for example node has (this is node, so it can it could always find other solutions) this pattern internally called "primordials" where it like pulls in a fresh pre-frozen copy of prototype methods, but its use of spread syntax is always vulnerable to user code breaking node itself and it's really painful to try and rewrite things to not use things like spread syntax to avoid this problem. As I said, node has other alternatives because it's an engine, but the same sort of thing happens in user code and like babel output and so on. It's a lot that has to be worked around. So this is a wonderful change.
```

meetings/2020-11/nov-17.md
```diff
@@ -0,0 +1,729 @@
...
...ally agree with one of the things on your slide show. I don't think we have to worry too much about someone making a package that does the bad thing. I think they'll do that anyway if they want to; and they probably won't regardless, but that said, I think it looks like there's more risks than having it that aren't on this slide. As you mentioned, people will reach for `.slice(-1)[0]`, or people will have to manually compute the index that they want. I think that's a risk. That's gross code that's easy to screw up that people do all over the place. Gross is of course subjective, but I hope that that's not a contentious subjective opinion. So I think it's really important to have it. Code points aren't enough to solve what people actually want; which is some form of grapheme clusters. I think that there's a very large amount of strings that aren't human sentences where surrogate pairs matter. There's all sorts of things like code enums, and plenty of ASCII strings in which code unit indexing is exactly what's desired, which is why the “slice” and “length minus 1” patterns are common. So I hope that we can continue including this, just as I said during the original stage 3 promotion.
```
ljharb:
```suggestion
JHD: I don't actually agree with one of the things on your slide show. I don't think we have to worry too much about someone making a package that does the bad thing. I think they'll do that anyway if they want to; and they probably won't regardless, but that said, I think it looks like there's more risks than having it that aren't on this slide. As you mentioned, people will reach for `.slice(-1)[0]`, or people will have to manually compute the index that they want. I think that's a risk. That's gross code that's easy to screw up that people do all over the place. Gross is of course subjective, but I hope that that's not a contentious subjective opinion. So I think it's really important to have it. Code points aren't enough to solve what people actually want; which is some form of grapheme clusters. I think that there's a very large amount of strings that aren't human sentences where surrogate pairs matter. There's all sorts of things like code enums, and plenty of ASCII strings in which code unit indexing is exactly what's desired, which is why the “slice” and “length minus 1” patterns are common. So I hope that we can continue including this, just as I said during the original stage 3 promotion.
```

meetings/2020-11/nov-18.md
```diff
@@ -0,0 +1,684 @@
...
+
+DRO:  Just a suggestion: if you're talking about making a picker or a list view for that area, generally speaking, my recommendation would be to create a new input type (e.g. `<input type=”timezone”>`) so that the control over what is shown is given to the browser/user rather than relying on the good faith of the page or app that is using this enumeration. That kind of been a design that's done for other things that have been similar to this in the past. One example that is similar to this would be like trying to show a list of supported fonts. Rather than having a enumeration API for all of the fonts that exists on the system you have a way to have the browser give me a font and let the user pick, giving them the control.
+
+JHD:  I'm in the queue to reply to that. DRO any of these things that need to be available in the browser also need to be available on the server where HTML generated user input is evaluated - so any DOM element is just not a solution to those use cases. I realize that with a SPA that's fully client-rendered that would be sufficient, but many apps don't opt for that for various reasons.
```
ljharb:
```suggestion
JHD: I'm in the queue to reply to that. DRO any of these things that need to be available in the browser also need to be available on the server where HTML generated user input is evaluated - so any DOM element is just not a solution to those use cases. I realize that with a SPA that's fully client-rendered that would be sufficient, but many apps don't opt for that for various reasons.
```
(in general, you could probably `s/([^ ]) +/$1 /g` the entire document; there's no reason two spaces need to appear next to each other except for indentation at the start of a line)

meetings/2020-11/nov-19.md
```diff
@@ -0,0 +1,666 @@
...
...d be some idea of a solution for method extraction. So I'm very excited that you're bringing this back because it offers that. The concern I have which I think is a stage one concern and something we don't have to decide today. But is this seems like this is doing a lot of things in order or your suggestion is that it's doing a lot of things to try to solve a few different problems the problem that problem that I'm very that I'm motivated to see solved is is the is the one. The slide that you have where you show the comparison between chaining and using `.call`. think it's one of your first couple slides that's the one I'm very interested in seeing go forward. I think that there's a lot of extra complexity that seems to overlap with the pipeline proposal. So I guess that my question is how do you see that overlap and In particular, (I'm trying not to focus too much on this suggested syntax), but in particular the form you're showing with like all the `.prototype` stuff that I think Mark alluded to as well. Yeah, I guess what are your thoughts on complexity concerns and overlap with the pipeline operator and if you already addressed this, I apologize for that, no need to recover it.
```
ljharb:
```suggestion
JHD: Hi, I'm very sorry if you went over this in the slide show; I came in a little late and I tried to skim through the notes and catch up on the slides. The original bind operator kind of had three purposes/functions/uses. When the pipeline proposal was brought up, the discussion we had in plenary was that pipeline satisfies two of the three and maybe I'm not being precise about the number "three", but the thing that pipeline did not offer was a robust form for method extraction. The thing that I brought up in the context of the pipeline operators that I would that it was very important to me that if since the committee had been deciding along with pipeline that the bind operator, was that we were sort of picking one over the other. In other words the sense in the committee at the time was that if the pipeline operator advanced, the bind operator would not, and the constraint that I raised was that I thought it important that before pipeline operators advance to stage 2 that there would be some idea of a solution for method extraction. So I'm very excited that you're bringing this back because it offers that. The concern I have I think is a stage 1 concern and something we don't have to decide today, but it seems like this is doing a lot of things to try to solve a few different problems. The problem that I'm very motivated to see solved is method extraction. The slide that you have where you show the comparison between chaining and using `.call` - that's the one I'm very interested in seeing go forward. I think that there's a lot of extra complexity that seems to overlap with the pipeline proposal. So I guess that my question is how do you see that overlap and In particular, (I'm trying not to focus too much on this suggested syntax), but in particular the form you're showing with like all the `.prototype` stuff that I think Mark alluded to as well. Yeah, I guess what are your thoughts on complexity concerns and overlap with the pipeline operator and if you already addressed this, I apologize for that, no need to re-cover it.
```

meetings/2020-11/nov-19.md
```diff
@@ -0,0 +1,666 @@
...
+
+SYG:  All right. So along the same line as Jordan was saying, maybe I have a slightly weaker understanding that the pipeline path and something like this would be mutually exclusive but it seems to certainly lean that way that it would be mutually mutually exclusive and I guess that's okay. So I'll State my high level concern which would be addressed if they were mutually exclusive. My high level concern is that if they were to exist there would be too many. There will be a proliferation of different syntax to do the same things with some of the use cases overlapping but not all and I think that would strictly work for readability. But of course if we only go with one of them then that's fine. So I guess the concrete concern is where the champions for the various pipeline operators are in the room. And when do we expect this question of which path do we take as a committee go forward, I guess from JHD’s point of view. This path has already been decided. It's your take from a previous consensus. Yeah, this was the explicit discussion that we had. I can probably dig up the notes when pipeline went to stage one. I see you thanks.
```
ljharb:
```suggestion
SYG:  All right. So along the same line as JHD was saying, maybe I have a slightly weaker understanding that the pipeline path and something like this would be mutually exclusive but it seems to certainly lean that way that it would be mutually mutually exclusive and I guess that's okay. So I'll State my high level concern which would be addressed if they were mutually exclusive. My high level concern is that if they were to exist there would be too many. There will be a proliferation of different syntax to do the same things with some of the use cases overlapping but not all and I think that would strictly work for readability. But of course if we only go with one of them then that's fine. So I guess the concrete concern is where the champions for the various pipeline operators are in the room. And when do we expect this question of which path do we take as a committee go forward, I guess from JHD’s point of view. This path has already been decided. It's your take from a previous consensus. Yeah, this was the explicit discussion that we had. I can probably dig up the notes when pipeline went to stage one. I see you thanks. 
```

meetings/2020-11/nov-19.md
```diff
@@ -0,0 +1,666 @@
...
+
+JHX: It seems like for the old proposal, the Champions do not want to push forward. So this is just another request that if this proposal, I would also like to ask if I can reuse the bind operator proposal. 
+
+RPR: So I think it's state stage 0 / proposals. Don't don't block the stage one, and we've already said that in stage one will figure out whether this conflicts with the pipeline as well. So I don't think this is a stage one concern but this proposal.
+
+RBN:  All right. That's all I wanted to make sure. I have no other concern. Thank you, specifically.
+
+JHD: I don't think we should be dictating which like I don't think it makes a difference. It's whether this is a new repo in a new entry on the proposal stable versus whether it reuses the existing one or replaces the existing one. I think that at the time when something advances to Stage 2 is when we should be explicit about which pre-staged two things are effectively inactive as a result of that stage to advancement. So not today.
```
ljharb:
```suggestion
JHD: I don't think we should be dictating which; I don't think it makes a difference whether this is a new repo in a new entry on the proposals table, versus whether it reuses the existing one or replaces the existing one. I think that at the time when something advances to Stage 2 is when we should be explicit about which pre-stage-2 things are effectively inactive as a result of that stage to advancement. So, not today.
```

meetings/2020-11/nov-19.md
```diff
@@ -0,0 +1,666 @@
...
+
+JHD: The first question is just for the sake of the notes. Can you please explain what the JSCIG is?
+
+YSV: JSCIG is the JavaScript China interest group. So we I believe it is composed of a number of tc39 members from China. So I believe that includes JHX and a few other people.
+
+JHD: I think automated updating a links in the proposals repository and elsewhere is a win, we should Just do it. And while it's awesome that it's being talked about with the committee, I hope that that's not something we require consensus for.
```
ljharb:
```suggestion
JHD: I think automated updating a links in the proposals repository and elsewhere is a win, we should just do it. And while it's awesome that it's being talked about with the committee, I hope that that's not something we require consensus for. 
```

meetings/2020-11/nov-19.md
```diff
@@ -0,0 +1,666 @@
...
+
+YSV: I believe I do address it because as I mentioned this is similar to our break time when we transition to the next speaker setting up their laptop. Nobody leaves the room and generally nobody leaves their chair. And additionally if we don't want to have the five minute break, which is an option. We don't have to have the five-minute break. In that case the fallback here is we check using GitHub action which will fail if the conclusions are not in the correct format.
+
+WH: Let's move on.
+
+MS: Isn't one minute enough?
+
+YSV: The time is arbitrary that I suggested. What I meant is a short period of time that's dedicated to making sure that that gets checked. Hopefully that answers your question 
+
+JHD:. yeah, just want to say I also think that having structured JSON metadata in the proposals repo that's used to generate the markdown in it and then consumed however is like is also a when we should also do that and hope that also wouldn't require consensus. That's a great idea
```
ljharb:
```suggestion
JHD: I also think that having structured JSON metadata in the proposals repo that's used to generate the markdown in it is also a great idea, and hope that also wouldn't require consensus.
```

meetings/2020-11/nov-19.md
```diff
@@ -0,0 +1,666 @@
...
...u can call it multiple times and every time you get a fresh copy to expect the resulting object immutable. I think the better analog is nodes require of a JSON file. As soon as you require a JSON file, that value is cached for the rest of the program and any mutations to it when - [audio drops] Okay Okay, so node require when require a JSON file that it caches it as a Singleton for the lifetime of the program. Mutations are visible to any user who looks at it after that, just as they would be with the mutable form of this proposal. In practice this is fine. Most people don't mutate anything and if they do it's because they want everyone else to see the mutations and if they want it frozen, they're welcome to do that and that would prevent anyone else from mutating it in the future. So I think that the user expectations here are both that it be mutable. And also I think that about a decade of people doing this in node has shown that there are no problems caused by like in terms of how it works - I'm not talking about like performance or anything, but there are no like ergonomics problems caused by having it be mutable. So I am very strongly prefer that this be mutable to match that.
```
ljharb:
```suggestion
JHD:(topic: "user expectation - JSON.parse is mutable in practice - this is fine") So as was pointed out to me and in chat after I put this on queue it to is probably not the the best analog because you can call it multiple times and every time you get a fresh copy to expect the resulting object immutable. I think the better analog is nodes require of a JSON file. As soon as you require a JSON file, that value is cached for the rest of the program and any mutations to it when - [audio drops] Okay Okay, so node require when require a JSON file that it caches it as a Singleton for the lifetime of the program. Mutations are visible to any user who looks at it after that, just as they would be with the mutable form of this proposal. In practice this is fine. Most people don't mutate anything and if they do it's because they want everyone else to see the mutations and if they want it frozen, they're welcome to do that and that would prevent anyone else from mutating it in the future. So I think that the user expectations here are both that it be mutable. And also I think that about a decade of people doing this in node has shown that there are no problems caused by like in terms of how it works - I'm not talking about like performance or anything, but there seem to be no ergonomics problems caused by having it be mutable. I very strongly prefer that this be mutable to match that.
```

ljharb (approved):

sffc (approved):

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
+IS: Okay, that's that's a good good catch. OK, so apparently we have a conflict with TC39’s new schedule which was presented just half an hour ago and the GA meeting. Yes, so we since it is difficult to move the general assembly meeting. So I would say that we have to move the TC39 meeting. It would be better the week beforeI know can become brings theOut to the general assembly meeting. 
+
+RPR: Alright, will review that now and make a swift change to the tc39 meeting them? Yeah. thanks Waldemar. Thank you very much. That was a very good catch.
+
+
+## ECMA262 status update
+Presenter: Kevin Gibbons (KG)
+
+- [spec](https://github.com/tc39/ecma262)
+- [slides]()
```
bakkot:
```suggestion
- [slides](https://docs.google.com/presentation/d/1MtgFNmwetns_mOWu0k6VKfwAuLrqom8GZdQhHna4WBE/edit)
```

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
...rray is detached at the beginning of the method call? And what this PR does is then tries to follow the spirit of what the process would be. So it first copies the spec text word for word step for step from the array dot prototype, then inserts a valid typedarray call at the start, then substitutes all the reads of the length property with reads of the array length internal slot. And this is highlighted because this is the part that has disagreement and is under-specified and in red. The has property checks because type of a stand-up holes and implementations don't call hasProperty. Notably the prose doesn't say anything about what you should do about the hasProperty checks. So the implementations kind of interpreted some way and did something. Luckily only three methods care about the HasProperty checks. So steps 1 to 3 here is sufficient to fully define in a compatible way for all the methods except these three methods: includes indexOf and lastIndexOf. Once a typed array type to raise the underlying buffer is detached, type 3 behaves as if they have no index to come properties. So all calls to the hasProperty, the hasProperty internal method will return false for its properties.
```
syg:
```suggestion
SYG: So this is another thing where in Ross's quest to get the spec to reflect reality about typed arrays, we found some other parts that were under specified and have some engine disagreement. So this is one of those follow-on PRs I'm presenting for Ross. So currently on TypedArray.prototype for the different TypedArrays, the first bullet point lists [on the slide] all the things that do not have algorithm steps. They only have prose. The only two methods that have algorithm steps are map and filter. So what the PR does is it tries to give everything algorithm steps, but we ran into some issues. So first of all what the prose currently says is the following three things. I'm paraphrasing of course, but they all basically say to implement the same thing as the Array.prototype counterparts except when you would read that length property, instead read the internal slot [[ArrayLength]] on the typed array, and at the start of every method call validate typed array basically all valid a type. All it does is it throws if the typed array is detached at the beginning of the method call. And what this PR does is then tries to follow the spirit of what the process would be. So it first copies the spec text word for word step for step from the Array.prototype, then inserts ValidateTypedArray call at the start, then substitutes all the reads of the length property with reads of the [[ArrayLength]] internal slot. And this is highlighted because this is the part that has disagreement and is under-specified and in red. The [[HasProperty]] checks because TypedArray doesn't have holes and implementations don't call [[HasProperty]]. Notably the prose doesn't say anything about what you should do about the [[HasProperty]] checks. So the implementations kind of interpreted some way and did something. Luckily only three methods care about the HasProperty checks. So steps 1 to 3 here is sufficient to fully define in a compatible way for all the methods except these three methods: includes indexOf and lastIndexOf. Once a typed array type to raise the underlying buffer is detached, type 3 behaves as if they have no index to come properties. So all calls to the hasProperty, the [[HasProperty]] internal method will return false for its properties.
```

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
...ly the same example as before but instead of detaching, the length is 0. Here the engines all agree. We all say that a sneakily truncated array behaves as if it includes undefined. So recall again, the only engine that agrees with arrays today is spider monkey. So the proposal is to have spidermonkey semantics, to align with what the array analogy does. I think the web compat risk is very low. I don't think people are writing code like this. It seems very unlikely to exist in the wild. IndexOf again. Here's two examples that kind of tries to sneakily detach a typed array and truncate a regular array. For indexOf, V8 is the one that aligns with the array behavior. meaning array index of and typedarray index of both have a hasProperty check. and because a detached typedarray behaves as if it has no own properties, it does not find the index of undefined.  And same thing for lastIndexOf in reverse. So again aligned with the semantics here which aligns with the array semantics. For spider monkey least I don't think this is controversial. There's a join method that can also sneakily detach or truncate the input array and array that's joined is as if it's array full amount of undefineds.
```
syg:
```suggestion
So first, I'm going to show you how we differ. suppose we have the following code. We make a new Uint8array and then we make a poison pill that sneakily detaches the typed array when you use it as a value and then we check if the typedarray includes the value undefined. and the algorithm steps of aray.prototype.includes where we copied to the algorithm steps from what it does is, it gets the the value of the second argument which is the start index to search for my thing like it does the value of conversion and stuff after it validates that es that this argument so the validation in this case. At first it's not detached, so it doesn't throw but then it detaches when it tries to coerce or convert. So what should this do currently? JSC throws because JSC kind of throws on all the detached stuff; SpiderMonkey doesn't, it returns true meaning for  sneakily detached; V8 returns false, saying the detached TypedArray does not include undefined or the proposal is to not aligned with you know, step by step what the array methods do but instead try to align with the spirit of what we interpret the spirit of the original Pros definition to try to mean which is aligned with the observable output of the array method. So what I think the observable behavior ought to be with the detached case is to be the same as the analog of setting the length of a regular array of zero. I consider that an analogous case to sneakily detaching typed array. I sneakily set the length of a regular rate to zero. So this is basically the same example as before but instead of detaching, the length is 0. Here the engines all agree. We all say that a sneakily truncated array behaves as if it includes undefined. So recall again, the only engine that agrees with arrays today is spider monkey. So the proposal is to have SpiderMonkey semantics, to align with what the array analogy does. I think the web compat risk is very low. I don't think people are writing code like this. It seems very unlikely to exist in the wild. IndexOf again. Here's two examples that kind of tries to sneakily detach a typed array and truncate a regular array. For indexOf, V8 is the one that aligns with the array behavior. meaning array index of and typedarray index of both have a hasProperty check. and because a detached typedarray behaves as if it has no own properties, it does not find the index of undefined.  And same thing for lastIndexOf in reverse. So again aligned with the semantics here which aligns with the array semantics. For spider monkey least I don't think this is controversial. There's a join method that can also sneakily detach or truncate the input array and array that's joined is as if it's array full amount of undefineds. 
```

meetings/2020-11/nov-16.md
```diff
@@ -0,0 +1,678 @@
...
+
+WH: You said that if you call includes on an empty array and you search for undefined it returns true, but it actually returns false.
+
+SYG: what if you if you search on empty array for undefined you get false you right you search an empty array in this fashion where you sneakily truncate the array? in the conversion of the second argument you initially when you called includes the array is not yet empty. Okay. Does that clarify? 
+
+WH: Yeah, I'm just trying to figure out what the regular array semantics are that you're trying to match here.
+
+SYG: What I'm matching is that in typedrrays when I first call?The typedaray rate it's not yet detached its buffer is not yet attached with attaches. IIt is the conversion the second argument.
```
syg:
```suggestion
SYG: What I'm matching is that in TypedArrays when I first call includes, the TypedArray isn't yet detached, it's still attached. It is the conversion the second argument that causes the TypedArray to be detached.
```

meetings/2020-11/nov-18.md
robpalme: Maybe there's a markdown linter that could enforce this?

PR #127, opened 2021-05-20
chore(delegates): add Attila Bartha

atikenny:


delegates.txt
```diff
@@ -26,6 +26,7 @@ Anne van Kesteren (AVK)
 Arnaud Le Hors (ALH)
 Ashley Hauck (AEH)
 Ashley Williams (AWS)
+Attila Bartha (AB)
```
ljharb: other than legacy examples, i believe we try to have all abbreviations be three letters.

ljharb (approved):

robpalme (approved):

PR #134, opened 2021-05-26
Add Travis to delegates list

travisleithead:
New delegate onboarding from Microsoft


robpalme (commented):
TLAs are usually of the form first-letter-of-firstname, first-letter-of-surname,  last-letter-of-surname

Would you be ok with TLD instead of TSL? It also has more gravitas imo.

robpalme (approved):

robpalme:
Will merge this as TLD next week unless there is an objection.

travisleithead:
Sure TLD is fine. Thanks!

PR #137, opened 2021-06-11
Add notes for 2021-05

robpalme:
ref: https://github.com/tc39/Reflector/issues/379
ref: https://github.com/tc39/Reflector/issues/367


ptomato (approved):
I reviewed them in the docs, assuming they're unchanged :smile:

PR #142, opened 2021-07-16
Add J. S. Choi to delegates list

js-choi:
New employee of Indiana University. CC: @samth.


gesa (approved):
ooh JSC is a most excellent TLA.

PR #144, opened 2021-08-03
Add July 2021 minutes

robpalme:
ref: https://github.com/tc39/Reflector/issues/390
ref: https://github.com/tc39/Reflector/issues/380


meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+BT: All right, the queue is empty.
+
+### Conclusion/Resolution
+No changes sought or made, general agreement on the status quo
+
+
+## Decorators update
+Presenter: Chris Garrett (CHG)
+
+- [proposal](https://github.com/tc39/proposal-decorators
```
Jxck: missing closing `)`

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+AKI: Thanks for that chip. That's a good point.
+
+SYG: So maybe I misunde the actual intention of this proposal. Is the proposal to change the process document or to kind of get us on the same page to try to stop using the word straw person?
```
gesa:
```suggestion
SYG: So maybe I misunderstand the actual intention of this proposal. Is the proposal to change the process document or to kind of get us on the same page to try to stop using the word straw person?
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+AKI: I think it's both.
+
+SYG: I see. It seems like it would be clear for the process document to perhaps just drop any mnemonic.and just, we just refer to Stage 0 through 4. As for getting us on the same page, that seems like a taller order but because I have nothing against "concept". Sorry, that was not very clear. I'm proposing that instead of spending time in plenary to debate whether "concept" is a good replacement for strawperson, we could just drop it from the process document.
```
gesa:
```suggestion
SYG: I see. It seems like it would be clear for the process document to perhaps just drop any mnemonic just refer to Stage 0 through 4. As for getting us on the same page, that seems like a taller order but because I have nothing against "concept". Sorry, that was not very clear. I'm proposing that instead of spending time in plenary to debate whether "concept" is a good replacement for strawperson, we could just drop it from the process document.
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+AKI:. Thank you so much. I will not be joining anybody for skiing because that's why I live in San Diego. I hate snow.
```
gesa, line 59: Could you do me a solid and scrub this from the commit history?

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+MM: Yeah. So, absolutely definitely interested in collaborating on this proposal and seeing it move forward. We cannot write a session that stays secure into the future without something like this.
+
+USA: one question from Greg, from the chat is Jordan are you looking for implementer concern?
+
+JHD: Yeah, I mean in particular, since the of the friction that Realms at that has been implemented concerned. So I want to I, particularly for that. Thank you Greg.
+
+AKI: Real quick. I just want to make sure that you take the time to throw. probably wouldn't want to talk. Technical things are not supposed to be on discourse, maybe try to reach out to people who weren't able to make this kind of very specific time zone Restricted meeting time looking for the set that same feedback of find out ahead of time. If there would be any concerns or if or if there would be any Issues that you should be prepared for before you get started. This could make sure to ask outside of this meeting. There's a lot of people who couldn't be here, who may or may not have opinions. I have no idea.
```
gesa: IDK how I both messed this one up and missed it on first review. Extremely false.

```suggestion
```

gesa (commented):
whups.

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+CP: We also have a polyfill that we have implemented a few months ago, we have test for that too. We actually validate one of Jack's concerns which was whether, or not the callable boundary was enough to be able to implement a membrane on top of it. So we have a very tiny Library, about 1kb, 1.5kb, very simple implementation, but a robust implementation of a membrane library. That allows you to do that kind of advanced use cases if that's what you need. Many of the use cases don't even need that, instead you just need a way to create a realm to evaluate code and maybe get access to some bits out of the realm. So with that in mind, we believe that we're ready for stage 3 and we are ready to open the debate. Questions?
+
+JHD:. So the first one I added was, I mean if this event since this membrane Library, so tiny, why not just support that directly in the language as opposed to making people add this easy to get wrong Library, even if it's small?
```
ljharb:
```suggestion
JHD:. So the first one I added was, I mean if this event since this membrane library is so tiny, why not just support that directly in the language as opposed to making people add this easy-to-get-wrong library, even if it's small?
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+CP: yeah, I want to add one thing to what, just, what marks said and is that the callable boundary? There is actually help a lot when it comes to having that membrane between two rooms and we have been working on this for quite some time and before they call Obama it was a thing and it was a common mistake when you implement a membrane to have issues with the separation and object from the other side. And so on, especially with errors and throwing errors between implications across the boundary. So it actually helps a lot to have that separation place.
+
+JHD: Okay, and then just a real quick and then I'll move on to the replies so you're saying that there is no. larger feature before you get to the points of disagreement that Mark described. Beyond this. Like there's no there's no hop between callable Realms and full membranes with a decided distortion mechanism.
```
ljharb:
```suggestion
JHD: Okay, and then just a real quick and then I'll move on to the replies so you're saying that there is no larger feature before you get to the points of disagreement that Mark described. Beyond this. Like there's no there's no hop between callable Realms and full membranes with a decided distortion mechanism?
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+CP: Yeah. So then we call this Y 2 Jordan, the terms that we use for that is integrity preserving, semantics of the callable boundary. So if you just cannot share that the object. 
+
+JHD: So just to clarify, again, Current the boundary API does not permit. Any way to get direct object object access access over here. He's taste test. And you're saying that's like safer because it just does what you want already and handles. All of the separation for you is that accurate correct?
```
ljharb:
```suggestion
JHD: So just to clarify, again, Current the boundary API does not permit any way to get direct object access access, and you're saying that's safer because it just does what you want already and handles all of the separation for you. is that accurate? 
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+CP: Yeah. So then we call this Y 2 Jordan, the terms that we use for that is integrity preserving, semantics of the callable boundary. So if you just cannot share that the object. 
+
+JHD: So just to clarify, again, Current the boundary API does not permit. Any way to get direct object object access access over here. He's taste test. And you're saying that's like safer because it just does what you want already and handles. All of the separation for you is that accurate correct? 
+
+JRL: Correct. We don't want to share objects. We only need the ability to run on other code without giving them access to our globals, the integrity boundary. 
+
+JDH: So just as a brief thought, experiment to make Sure, I have the understanding correct. If the old Realms API had some sort of Constructor option that like opted you into this mode, such that you didn't have to worry about it. Would that achieve the same goal for you?
```
ljharb:
```suggestion
JHD: So just as a brief thought experiment to make sure I have the understanding correct: if the old Realms API had some sort of constructor option that like opted you into this mode, such that you didn't have to worry about it, would that achieve the same goal for you? 
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+JHD: So to be cleared an the at the time it was brought to stage two. It's solved the shared object use cases also, so callable boundary around is effectively a brand new proposal that happens to solve some of the use cases of the proposal that went to stage two. So it's I think it's an it's not really fair to characterize this as like, oh well, these are just two overlapping proposals because, like, essentially the stage 2 proposal was kind of replaced very recently replaced with a One that does not solve all of the use cases that the stage 2 proposal solved. The fact that the champion group at only has a subset of those use cases, doesn't also negate that the stage 2 presumption. We all had was that all of those use cases were solved and that's what my next sort of queue item is around.
```
ljharb:
```suggestion
JHD: So to be clear, at the time it was brought to stage 2, it solved the shared object use cases also, so callable boundary is effectively a brand new proposal that happens to solve some of the use cases of the proposal that went to stage 2. So it's I think it's not really fair to characterize this as like, "oh well, these are just two overlapping proposals" because essentially the stage 2 proposal was kind of replaced  - very recently - with one that does not solve all of the use cases that the stage 2 proposal solved. The fact that the champion group only has a subset of those use cases, doesn't also negate that the stage 2 presumption we all had was that all of those use cases were solved and that's what my next sort of queue item is about. 
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
...allable the boundary in is enforcing isolation. And if the isolation is optional, then it's not something that needed to be enforced by an inescapable mechanism in which case building it into the platform is, is with note is no longer solving a fundamental problem if you make it optional, because the only fundamental difference between providing at the user level and providing it in the platform is that the is that we can make it non-optional by putting to into the platforms. So, once again, historically I think it's the the way we ended up here is that some of the people objectioning to the nature of the stage 2 proposal wanted to make the isolation mandatory. So I think I have trouble characterizing this as well. If we want to make it optional, can add an option to make it optional in the future. I'm uncomfortable with that because if we wanted to make it optional, then I agree that we shouldn't introduce the callable boundary in the first place. But, what we found was we were deadlocked on moving it forward, whereas the callable boundary meets the use cases that most of us have in mind, That would be happy either way. And satisfies those who want to make the isolation mandatory.
```
ljharb:
```suggestion
MM: Yeah. Historically, JHD is correct, The Realms without the callable boundary. Certainly enables the call, the boundary, to be built at the user level. the thing that solved by building the callable the boundary in is enforcing isolation. And if the isolation is optional, then it's not something that needed to be enforced by an inescapable mechanism in which case building it into the platform is, is with note is no longer solving a fundamental problem if you make it optional, because the only fundamental difference between providing at the user level and providing it in the platform is that the is that we can make it non-optional by putting to into the platforms. So, once again, historically I think it's the the way we ended up here is that some of the people objectioning to the nature of the stage 2 proposal wanted to make the isolation mandatory. So I think I have trouble characterizing this as well. If we want to make it optional, can add an option to make it optional in the future. I'm uncomfortable with that because if we wanted to make it optional, then I agree that we shouldn't introduce the callable boundary in the first place. But, what we found was we were deadlocked on moving it forward, whereas the callable boundary meets the use cases that most of us have in mind, That would be happy either way. And satisfies those who want to make the isolation mandatory.
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
...is that 90% of the people who want solve this problem in this room, and that's a ballpark number. Some will, have their use cases, men and will. And we'll have much lesser much reduced forms of motivation to advance things, and the people who have their who have created, this sort of Deadlock and forced our alternative. Forced to callable Realms. boundary callable boundary Realms as an alternative. Will not have ever needed to come into this room and argue their points directly. They'll they have only done it by proxy and that is very frustrating because It doesn't matter. Solely what the Use cases the champion group has the proposal as it entered stage to solve a set of use cases. The champion group may only cover a subset of those, but all of those use cases are what like met the stage Two criteria we expect. So for the solution into the language, we agree. Those are these cases We want to solve this approach is effectively magically creating a new proposal that's already at stage 2, somehow. And asking for it to go to stage 3 while simultaneously magically discarding the existing proposal and I'm saying that in the sense that like some of those use cases are no longer achievable
```
ljharb:
```suggestion
JHD: So, Okay, yeah, thank you. As Mark has pointed out, there have been some blocking objections by folks who are represented by people in this room, but who have not debated these objections directly in this room. It's my understanding that this is basically saying the capability (that browsers and node users already have) should not be something we permit in a new API for "reasons". All the arguments I've heard are very good, compelling arguments for what the default should be. The default should be what Justin and AMP needs - which is the safer thing, where you can't escape the abstraction. That's a very compelling argument, but none of those arguments preclude the capability, and there are use cases that require the capability. I specifically had these cases where I need to get objects with internal slots. A Proxy to it is not sufficient for built-ins like that. I could probably write a wrapper that like, detects an incoming thing with slots and reconstructs the thing with slots coming out, but that's going to very difficult and very easy to get wrong. So I guess it's if I think it's like my personal feeling is that 90% of the people who want solve this problem in this room, and that's a ballpark number. Some will have their use cases met, and we'll have much lesser, much reduced forms of motivation to advance things, and the people who have created this sort of deadlock and forced callable boundary Realms as an alternative will not have ever needed to come into this room and argue their points directly. They'll they have only done it by proxy, and that is very frustrating because It doesn't solely matter what use cases the champion group has - the proposal as it entered stage 2 solved a set of use cases. The champion group may only cover a subset of those, but all of those use cases are what were met. The stage 2 criteria means we expect the solution to go into the language, we agree those are these cases we want to solve. This approach is effectively magically creating a new proposal that's already at stage 2, somehow - and asking for it to go to stage 3 while simultaneously magically discarding the existing proposal. I'm saying that in the sense that some of those use cases are no longer achievable.
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+JHD: I'll try and speak slowly maybe not too slowly, had four minutes left shame. So we basically switched from the “only has the foot gun / capability” version to an “Only lacks the foot gun / capability” version. The other option that like, you know, that I would be more interested in is something that defaults to the safer, more restrictive one and allows opting in to the Wider, more powerful, but, you know, easier to screw up version. To me, that would be that would solve all of the use cases presented, my guess is - I know without having that version available and regathering, the evidence that Shu was talking about. I don't know how to confirm that guess. But my guess is that if we named everything, appropriately, whatever that means that we would be able to convey the risk and reduce the foot gun, likelihood that. So that's what I'm interested in seeing.
```
ljharb:
```suggestion
JHD: So we basically switched from the “only has the foot gun / capability” version to an “Only lacks the foot gun / capability” version. The other option that I would be more interested in is something that defaults to the safer, more restrictive one and allows opting in to the wider, more powerful, but easier to screw up version. To me, that would solve all of the use cases presented. I know without having that version available and regathering the evidence that SYG was talking about. I don't know how to confirm that guess. But my guess is that if we named everything appropriately, whatever that means, that we would be able to convey the risk and reduce the footgun likelihood. So that's what I'm interested in seeing.
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+SYG: I think Jordan would if I read you correctly, you were directly addressing Chrome in the good faith path forward. Not not that the champion group right now, the champion group. 
+
+JHD:That's right.
```
ljharb:
```suggestion
JHD: That's right.
```

meetings/2021-07/july-13.md
```diff
@@ -0,0 +1,512 @@
...
+
+JHD: So maybe another day I'll discuss that. 
+
+USA: Okay. Wait, we're at the top of the time box. Thank you everyone for the discussion. Do we want to continue this at a later time as an extension? 
+
+CP: Well, first I would like to ask is if there is any objection for stage 3 at this point.
+
+JHD: There needs to be discussed out before we get to that question. 
+
+SYJ: Is there anything that we can work through to we re discussed it? And what are the available slots for continue this discussion?
```
ljharb:
```suggestion
SYG: Is there anything that we can work through to we re discussed it? And what are the available slots for continue this discussion? 
```
(i *think* this is the correct speaker)

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
... 4 feature. However, if you would like to determine if an object has a private field, currently you have to use try-catch. This proposal gives you a straightforward boolean check of whether a private field is in object and that includes fields, private methods, and private accessors. The proposal stage 3, and I have all of these check marks here to measure progress toward stage four. So, as of this morning, Firefox version 90 is available Chrome. 91 has been available for a while. With this feature. The features also in webkit nightly and should be in the next Safari technology preview, which I believe will make it into the next full version of safari. But of course, Apple never discloses that before it's released. It's in node as a version 16.4.0. And it's also will be default. It's been in Babel for a while, it will be on by default with the end preset as soon as someone releases the package and merges the pr, which will happen after this feature gets stage the pull request of the spec exists and is approved. Test 262 test also have been landed. And so based on all of that I believe it clearly meets the criteria for stage four and want to ask the committee for consensus for that.
```
ljharb:
```suggestion
JHD: So, as a quick refresher, class fields is a stage 4 feature. However, if you would like to determine if an object has a private field, currently you have to use try-catch. This proposal gives you a straightforward boolean check of whether a private field is in an object and that includes fields, private methods, and private accessors. The proposal is stage 3, and I have all of these checkmarks here to measure progress toward stage 4. So, as of this morning, Firefox version 90 is available, and Chrome 91 has been available for a while, with this feature. The feature is also in Webkit Nightly and should be in the next Safari Technology Preview, which I believe will make it into the next full version of Safari. (But of course, Apple never discloses that before it's released). It's in node as of v16.4.0. It's been in Babel for a while; it will be on by default with the env preset as soon as someone releases the package and merges the pr, which will happen after this feature gets stage 4. The pull request to the spec exists and is approved. test262 tests also have been landed. And so based on all of that I believe it clearly meets the criteria for stage 4 and want to ask the committee for consensus for that.
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+SYG: Yeah, I think I agree that these are orthogonal concerns.
+
+JHD: I don't think there's any way we can phrase this that won't be relatively trivial to bypass. If we have a list, then someone will come up with a new way to mean JavaScript, right? Like JS, JavaScript, jscript. I don't know. Someone will come up with something, but then like, if it's if we do the like prose option, I mean Source, Testament, text module records, I can make my own kind of module record as a host that is a source text module record but with some optional trivial difference, which means effectively, includes all sorts text module records and then I can bypass this assertion entirely and still be in compliance. So, like, I think that on some level, I don't think any host is likely to do this, and if they do they wouldn't be able to leverage almost all the code out there that's using no assertions to bring in JavaScript. So like it kind of seems - it seems very unlikely to me that without a restriction this would become an interop concern because I can't see enough people using such a host with that assertion. So it kind of seems like the way things are is ideal.
```
ljharb:
```suggestion
JHD: I don't think there's any way we can phrase this that won't be relatively trivial to bypass. If we have a list, then someone will come up with a new way to mean JavaScript, right? Like JS, JavaScript, jscript. I don't know. Someone will come up with something, but then like, if it's if we do the like prose option, I mean Source Text Module Records, I can make my own kind of module record as a host that is a source text module record but with some optional trivial difference, which means effectively, includes all sorts text module records and then I can bypass this assertion entirely and still be in compliance. So, like, I think that on some level, I don't think any host is likely to do this, and if they do they wouldn't be able to leverage almost all the code out there that's using no assertions to bring in JavaScript. So like it kind of seems - it seems very unlikely to me that without a restriction this would become an interop concern because I can't see enough people using such a host with that assertion. So it kind of seems like the way things are is ideal.
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+MM: And if the getter and setter is applied to anything other than an instance, it throws?
+
+DDC: it should. If that's not in the spec then I should definitely add it. Any other questions?
+
+JHD: Yeah, it was a couple slides back where you're showing the decorator function interface. Yeah, here you go. I was just curious if the context access set function, can that ever fail, or it was that only there on a private field and so like it should never fail.
+
+DDC: If it is called on an instance of the class, because the class should always have that, it should never fail.
+
+JHD: It should never fail, but if it's called on the instead something else, so it'll either throw because the field isn't on the receiver. The receiver or it will succeed. Is that correct?
```
ljharb:
```suggestion
JHD: It should never fail, but if it's called instead on something else, so it'll either throw because the field isn't on the receiver, or it will succeed. Is that correct?
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+JHD: So two questions. The first was with the auto accessor feature that you described here, would this a viable standalone proposal, that could move independently as long as it was faster than independently of decorators and have value on its own, or do you think it only has value with decorators?
+
+DDC: I think it could move independently on its own. I definitely think that there are a lot of existing decorator use cases that depend on the ability to intercept access on fields. So either this proposal would have to kind of really be concurrent or ideally land first, if it were to be a separate proposal.
+
+JHD :Okay. And then the second question was with your "similar to". I was curious how faithful that sugar was. In other words, will the getter throw if it's called an object that isn't it? Like the didn't run through the constructor? You know like where as if it was a private field like your line 6 through 16 here will throw if the receiver lacks that private field. So I'm wondering if the auto accessors works the same. Then if so, how would I determine in advance without a try/catch if the accessor is going to throw or not?
```
ljharb:
```suggestion
JHD: Okay. And then the second question was with your "similar to". I was curious how faithful that sugar was. In other words, will the getter throw if it's called an object that isn't it? Like the didn't run through the constructor? You know like where as if it was a private field like your line 6 through 16 here will throw if the receiver lacks that private field. So I'm wondering if the auto accessors works the same. Then if so, how would I determine in advance without a try/catch if the accessor is going to throw or not?
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+DDC: That is a great question. The second one, rather. The first one is the answer is it will throw, I believe, because in the spec we back the field with a private slot, it uses all the same mechanisms as a private field. For the second question, I'm not too sure. I think you would have to use a try/catch. I'd have to think on that or what it means that like General concern, 
+
+JHD: there is a like address the proposal I had a few ago that just it stage 4 is like I wanted to determine if something is going to throw or not with the in keyword so that works for private Fields. But in this case because it's sugar for it, there isn't Special way to refer to it that would work within and tell you if it's going to throw or not. Because if did some value in an object, it'll just tell you if the accessors, you know, if the property is successful there, right?
```
ljharb:
```suggestion
JHD: there is the proposal I discussed recently that just got stage 4 is that I wanted to determine if something is going to throw or not with the `in` keyword, so that works for private fields now. But in this case because it's sugar for it, there isn't a special way to refer to it that would work within and tell you if it's going to throw or not. Because if did some value in an object, it'll just tell you if the accessors, you know, if the property is successful there, right?
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+JRL: `Date.now()` is a method that you call and having `Temporal.now` be lowercase makes me think it's supposed to be a callable function that's going to give me like date, time, or something in the Temporal object. Making it a capital namespace is going to prevent me from trying to just call it randomly.
+
+JHD: This is is a counter to my own constructor comment as well, but if you think it's a thing that it's not, and you try to use it that way, and then it immediately tells you it's not a function, you can't do something. If it's not a function, and you can't call it, you're not going to be confused anymore. So I don't know if the confusion argument really pushes in either direction very strongly.
```
ljharb:
```suggestion
JHD: This is a counter to my own constructor comment as well, but if you think it's a thing that it's not, and you try to use it that way, and then it immediately tells you it's not a function, you can't do something. If it's not a function, and you can't call it, you're not going to be confused anymore. So I don't know if the confusion argument really pushes in either direction very strongly.
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+SFC: I have two queue entries. First is, a Stage 0 proposal is not really formalized until it reaches Stage 1; it's not really even a proposal, it's almost more like a suggestion, because it's not really a proposal until it actually reaches Stage 1. So like if we were to keep the English names, I feel like Stage 1 being called "concept", Stage 0 being called something like "suggestion" or something like that.
+
+JHD: It's a proposal when a TC39 delegate champions it and plans to present it, that's usually the bar we use and that has been useful for things that did not make it to stage one, because they have come back again and they're still useful as reference points. So I think it's still useful to list stage 0.
```
ljharb:
```suggestion
JHD: It's a proposal when a TC39 delegate champions it and plans to present it, that's usually the bar we use and that has been useful for things that did not make it to stage 1, because they have come back again and they're still useful as reference points. So I think it's still useful to list stage 0.
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+AKI: Can I just repeat so I understand properly and in case other people are also not understanding? What you're saying is keeping stage zero on the process document is valuable because even though we're not committed to it, we can track that somebody did the work even if it never made it into the proper staging process. Is that what you're saying Jordan?
+
+JHD: Yeah. I mean I'm saying that the having a stage zero, having a perp repo for a proposal that is still in stage 0 but that has has or will get committee times as useful as a place to concentrate attention and track things, sure.
```
ljharb:
```suggestion
JHD: Yeah. I mean I'm saying that having a repo for a proposal that is still in stage 0 but that has or will get committee time is useful as a place to concentrate attention and track things, sure.
```

meetings/2021-07/july-14.md
```diff
@@ -0,0 +1,1028 @@
...
+
+JRL: So, I think with agreement that the Prototype object has to be changed to use object.create(null), or we have to use a Map, or a solution that does both. I think that's an agreement. There's also holes, which I think we still need to discuss but we can do that at stage one, and naming, which I think we're in agreement that just normal "groupBy" is fine. I think that is what I'm proposing to go to the stage one, that agreement
+
+JHD: I completely agree with the point making but I also think if we're going to consider both guilt or reject and groupBy separately, then they need to have two separate repositories for like Clarity and organization and whe named those proposals and what stage we decide there is a separate thing. But like, I think it's useful to have a specific repo that talks about group. I The problems at solving for the dresses, the Prototype thing and so on. And and I. So I think because of that, we need, it's prudent for us to separate it. Whether that's goes at stage one or two.
```
ljharb:
```suggestion
JHD: I completely agree with the point making but I also think if we're going to consider both filterReject and groupBy separately, then they need to have two separate repositories for clarity and organization, and what we name those proposals and what stage we decide they are is a separate thing. But I think it's useful to have a specific repo that talks about groupBy, the problems it is solving or addresses, the prototype thing, and so on. So I think because of that, we need, it's prudent for us to separate it. Whether that goes at stage 1 or 2.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+WH: I'm trying to understand your slide. The `fibonacciSequence` function returns an array from which Realm?
+
+CP: It is an array-like, and let us assume that it is from the realm that is executing this code on, which is not the other realm. Let's assume that for the sake of simplicity.
+
+JHD: I mean, the specific potential issue here is that this will return an array whose `.constructor` is `otherRealmGlobalThis.Array`. That's not `Array`. But this is only an issue If somebody's using instance of on arrays, which we've all been trained not to do for 10 years,
```
ljharb:
```suggestion
JHD: I mean, the specific potential issue here is that this will return an array whose `.constructor` is `otherRealmGlobalThis.Array`. That's not `Array`. But this is only an issue If somebody's using `instanceof` on arrays, which we've all been trained not to do for 10 years.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JHD: Arrays are kind of a unique example here, obviously, because we've all been using a raid on his array for a long time and arrays array. That array is proxies. So it's like one example but we're...
```
ljharb:
```suggestion
JHD: Arrays are kind of a unique example here, obviously, because we've all been using `Array.isArray` for a long time and Array.isArray tunnels through proxies. So it's like one example but we're...
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
...happens with those use cases. And I've certainly seen some examples like you've shown and heard some arguments that it's easy - if you have direct object access it is easy to get things wrong and that is certainly true. And it is certainly likely that there are alternative approaches than direct object access to another realm that would achieve the goals that I'm concerned about in ways that are harder to get wrong. I'm concerned that if this advances to stage 3 with this new form that has a restricted - like a subset of use cases met, that the sort of implied consensus that we also want to solve those other problems. Even if everyone may not have agreed with that like it would I would say implied is kind of gone. And that means that if I want to try and solve that use case, not only do I have to start from scratch with a new proposal, but I also have to come up with a mechanism that a lot of very experienced and intelligent people have not come up with, and admittedly weren't specifically trying to come up with, but it just kind of it's very strange to me that within stage to the shape of a solution would change so drastically such that it's almost a completely different proposal.
```
ljharb:
```suggestion
LEO: So JHD, are you feeling about it? I know you bring the main objection so far.

JHD: I have a few classes of concerns. So one is that this feels like a really weird and magic API for me. It's a very clever solution to the problem and I hear that it achieves the goals that people who want membranes have because it lets them use a membrane library still, but then it gives you better results with it and you know, takes more of the complexity and puts it in the platform. I get that. And I hear that AMP has a use case for which membranes work, but are overkill and that this all works in a more granular way. But this does not solve all of the use cases that this proposal went into Stage 2 with a solution for, regardless of what the original stated motivation was. That was the assumption at least that I was under, and I'm assuming that everyone else was under, when it went into stage 2. Then the other class of concern is separate from the strangeness of this new mechanism: the other concern is what then happens with those use cases. And I've certainly seen some examples like you've shown and heard some arguments that it's easy - if you have direct object access it is easy to get things wrong and that is certainly true. And it is certainly likely that there are alternative approaches than direct object access to another realm that would achieve the goals that I'm concerned about in ways that are harder to get wrong. I'm concerned that if this advances to stage 3 with this new form that has a restricted - like a subset of use cases met, that the sort of implied consensus that we also want to solve those other problems. Even if everyone may not have agreed with that like it would I would say implied is kind of gone. And that means that if I want to try and solve that use case, not only do I have to start from scratch with a new proposal, but I also have to come up with a mechanism that a lot of very experienced and intelligent people have not come up with, and admittedly weren't specifically trying to come up with, but it just kind of it's very strange to me that within stage 2 the shape of a solution would change so drastically such that it's almost a completely different proposal.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+CP: Can you be more specific about what those use cases you feel that this is not solving? assuming that obviously the one that we just discussed, we're saying doesn't solve that because that's problematic?
+
+JHD: Your example addresses certainly a big class of the use cases they have which is wanting to reduce the footprint of what I have to cache as first run code, such that I can then write robust code to write code that will run robustly later despite future modification of globalThis and the associated things. I hear the feedback that it's difficult to write that example correctly, but it's entirely possible to write it correctly. And the wrapper that I would write would construct a realm for myself, it would lock it down, maybe even with ses, but it would lock it down in some ways, and then it would I could even find functions and wrap them and like switch the prototypes and so on, there's there's lots of ways that it can be. Be constructed to be so that for my purposes, it will be correct. I'm not claiming that's easy. And the word that most people will fail,
```
ljharb:
```suggestion
JHD: Your example addresses certainly a big class of the use cases they have which is wanting to reduce the footprint of what I have to cache as first run code, such that I can then write robust code to write code that will run robustly later despite future modification of globalThis and the associated things. I hear the feedback that it's difficult to write that example correctly, but it's entirely possible to write it correctly. And the wrapper that I would write would construct a realm for myself, it would lock it down, maybe even with ses, but it would lock it down in some ways, and then it would I could even find functions and wrap them and like switch the prototypes and so on, there's there's lots of ways that it can be. Be constructed to be so that for my purposes, it will be correct. I'm not claiming that's easy.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JHD: that's that's not the problem I'm trying to solve, okay. So I think it is impossible to avoid having that problem. Once once you don't run first, all bets off. 
+
+MM: Good. 
+
+CP: Even the Realms can...
+
+JHD: Yeah you can you can modify because the the way into Realms which is pasta but you're also it's about my concern is about reducing the footprint scope of what I have to cash or hold onto that first run. So that it's all available later and the original Realms gives me one thing I have to grab this currently, I have a library that has 15 million downloads a week that I use for all my shims, that grabs everything and I have to explicitly, add each thing I need like, and then all of the shims have to like yeah it's just it's a it's a convenience and kind of scalability issue whereas Whereas, with the realm approach, I can it and of one and done to some degree.
```
ljharb:
```suggestion
JHD: My concern is about reducing the footprint and scope of what I have to cache or hold onto in that first-run code so that it's all available later, and the original Realms gives me one thing I have to grab. Currently, I have a library that has 15 million downloads a week that I use for all my shims, that grabs everything and I have to explicitly, add each thing I need, and then all of the shims have to load them all. It's a convenience and kind of scalability issue whereas, with the realm approach, I can do it "one and done" to some degree. 
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+MM: Good. 
+
+CP: Even the Realms can...
+
+JHD: Yeah you can you can modify because the the way into Realms which is pasta but you're also it's about my concern is about reducing the footprint scope of what I have to cash or hold onto that first run. So that it's all available later and the original Realms gives me one thing I have to grab this currently, I have a library that has 15 million downloads a week that I use for all my shims, that grabs everything and I have to explicitly, add each thing I need like, and then all of the shims have to like yeah it's just it's a it's a convenience and kind of scalability issue whereas Whereas, with the realm approach, I can it and of one and done to some degree. 
+
+CP: Well, one thing that we want to mention about this particular thing, Jordan is have done some it serves around these as well, in terms of birth, for four months. And if you grab that global object and you use all these do notations to access the things that you want when you eat it, you're going to pay the penalty as well. So that's another thing that at least another reason, why we catch everything because...
```
ljharb, line 503: i have no idea what this comment about birth, 4 months, and eating it, actually was. @fudco?

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JHD: I am a hundred percent unconcerned with performance until I have a hundred percent insured correctness. correctness. I will happy to cross that bridge when I come to it.
```
ljharb:
```suggestion
JHD: I am 100% unconcerned with performance until I have 100% ensured correctness. I will be happy to cross that bridge when I come to it.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JHD: You basically just described my get intrinsics tiebreak. So yeah, I mean, you're right, I don't need all of rounds for that. Use case, I need some, I'd like something for that, use case and Realms would have solved it. I'll be it in a way that requires careful work to do correctly.
```
ljharb:
```suggestion
JHD: You basically just described my `get-intrinsics` implementation. So yeah, I mean, you're right, I don't need all of Realms for that use case. I need something for that use case, and Realms would have solved it - all be it in a way that requires careful work to do correctly. 
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JHD: You basically just described my get intrinsics tiebreak. So yeah, I mean, you're right, I don't need all of rounds for that. Use case, I need some, I'd like something for that, use case and Realms would have solved it. I'll be it in a way that requires careful work to do correctly. 
+
+MM: So would you would you agree that something that gave you all of the original intrinsics from your own realm? When I say original, I don't mean unfortunately Which not I mean of course. 
+
+JHD: Analyzing at the moment that I've first, like, run my code.
```
ljharb:
```suggestion
JHD: I mean only what is available at the moment that I've first run my code. 
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+USA: We have some things on the queue just to give some status. There's been some chat the DC 39, delegates channel, one of the things that I caffeine has mentioned And that I think it's important if Kevin wants to bring it here but, also met you has some that, I think it's interesting for these days. 
+
+MAH: Yeah, I think similar idea, it seems that what Jordan is trying to do is get to the original intrinsics. And there are multiple options. One is a imperative API that you can call that's existing right now. Another one could be to re-explore the standard modules where you would be able to access those original intinsics things by importing them.
+
+JHD: that has the same problem as built-in modules themselves, which is unless I can synchronously access them in scripts, it is not sufficient.
+
+MAH: so, you You want them not only in
+
+JHD: just like levels, I want to be able to synchronously. Get them everywhere in JavaScript. Yeah, that's what that feature is here about which module system I'm using in nor should the future,
```
ljharb:
```suggestion
JHD: just like globals, I'd want to be able to synchronously get them everywhere in JavaScript.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
... think that seems a reasonable path forward under one - "condition" is the wrong word because that implies leverage but under the one, like in one circumstance, which is that the "direct object access for Realms" path. It does not appear to me that there will ever that will ever be allowed unless the current membership of TC39 completely turns over are so. It's, it would be kind of weird to ask for consensus on stage for stage one on a proposal or a problem that nothing was written up and put on the agenda before. right now. So it'd be nice if that could be granted, But that alone isn't sufficient because the real piece is, I would kind of need to know if any like, in the same way as when realms, Went to stage two. There was, it was, I think clearly telegraphed that there were these potential blocking concerns for the direct object access approach, right? Is there anyone in the room who has any reason to believe that such a proposal modulo, you know, dealing with various sorts of concerns that proposals run across, but that such a proposal would, would not have a path toward stage 4. Does that make sense? What I'm asking, I'm trying to figure out if it's worth me investing my time.
```
ljharb:
```suggestion
JHD: so, the queue is empty. I think that seems a reasonable path forward under one - "condition" is the wrong word because that implies leverage but under one circumstance, which is that the "direct object access for Realms" path does not appear to me that there will ever that will ever be allowed unless the current membership of TC39 completely turns over. It would be kind of weird to ask for consensus on for stage 1 on a proposal or a problem that nothing was written up and put on the agenda before right now. So it'd be nice if that could be granted, but that alone isn't sufficient because the real piece is, I would need to know - in the same way as when realms went to stage 2, it was clearly telegraphed that there were these potential blocking concerns for the direct object access approach, right? Is there anyone in the room who has any reason to believe that such a proposal modulo, you know, dealing with various sorts of concerns that proposals run across, but that such a proposal would, would not have a path toward stage 4. Does that make sense? What I'm asking, I'm trying to figure out if it's worth me investing my time.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+MM: Write the proposal, you're referring to is get in transits proposal. We Sure, yes. Okay, Good question.
+
+USA: Is it something that you'd like a temperature check on. 
+
+JHD: I mean, a temperature check seems nice, I think specifically if it would, even though Everyone is always allowed at any time to say, I have an objection or I have a constraint, even though I never mentioned it before, it would feel really bad if I put in time on a proposal like this and then came back in six months, a year, two months, whatever, and was surprised with a sudden constraint, it's happened to me once before it is not a fun thing to do and it's very demotivating. So I really just want to make sure that if there's any possibility that any one, could guess that, that might happen that I could get some hints about it now so that I know whether it's worth putting in my time, this is not really a technical or procedural question, I'm sort of asking you a personal question in somebody. Make sure that I'm not going to about to waste my time.
```
ljharb:
```suggestion
JHD: I mean, a temperature check seems nice, I think specifically if it would, even though everyone is always allowed at any time to say, "I have an objection" or "I have a constraint, even though I never mentioned it before", it would feel really bad if I put in time on a proposal like this and then came back in six months, a year, two months, whatever, and was surprised with a sudden constraint. It's happened to me once before, and it is not a fun thing to do and it's very demotivating. So I really just want to make sure that if there's any possibility that any one, could guess that, that might happen that I could get some hints about it now so that I know whether it's worth putting in my time, this is not really a technical or procedural question, I'm sort of asking you a personal question to try to make sure that I'm not about to waste my time.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JHD: No, in that like in that if somebody has simply mutated the object and I'm asking for the object, I want the object by identity, and so, it wouldn't any sense to me to have like a snapshot of it. It's more akin to like, if someone, if I cache getIntrinsic and then later, somebody says math.pi equals 3. And I get the math.Pi intrinsic. It will give me the value of pi.
+
+SYG: Yes, I think that is unproblematic. What is problematic for memory constraints especially is if we not have to give you back the identity they argue original object with the same identity but the old original actual original object in which case, you know that directly contradicts the need for identity. 
+
+JHD: If you do array.prototype.slice equals 3. And I with my cache to get intrinsic and I get the intrinsic for a rate of prototype that slice, or sorry. Yeah, Rita project that slice. Wait, hold on. If change the object, let's say the object prototype object.to something else. That's empty. And then I asked for object dot prototype dottostring. I need to get the original toString function.
```
ljharb:
```suggestion
JHD: If change the object, let's say the `Object.prototype` object to something else that's empty, and then I asked for `Object.prototype.toString`, I need to get the "original" `toString` function. 
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+  - Engines (at least V8, Node, FF) agree **not** to ship unflagged until after bikeshedding
+
+
+## `getOriginals` for stage 1
+Presenter: Jordan Harband (JHD)
+
+
+USA: Yeah. And next up we have Jordan couldn't do you want to ask for stage 1? Yeah. 
+
+JHD: So even though I don't yet have a repository, we sort of sketched it out here. I would be happy make one tomorrow after I get some sleep. It's is there any reason that could not be stage one already? I would be more comfortable with sleeping more.
```
ljharb:
```suggestion
JHD: So even though I don't yet have a repository, we sort of sketched it out here. I would be happy to make one tomorrow after I get some sleep. is there any reason that could not be stage 1 already?
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+Presenter: Jordan Harband (JHD)
+
+
+USA: Yeah. And next up we have Jordan couldn't do you want to ask for stage 1? Yeah. 
+
+JHD: So even though I don't yet have a repository, we sort of sketched it out here. I would be happy make one tomorrow after I get some sleep. It's is there any reason that could not be stage one already? I would be more comfortable with sleeping more. 
+
+SYG: So, I don't want to slow down your word, I don't want them walking, so let me hear from you. Do you think that if we wait for an official explain it, I'm proposing everything through grants X3 would that slow down velocity for you?
+
+JHD: No, it's just a nice signal. I think stage one is a problem statement, right? So the specifics of the solution don't need to be explained or understood just yet but - like I'm happy to write up that explainer, it's more like, if the problem is unclear to someone such that they are not willing to or not comfortable, making it stage one without that explainer, then obviously would need to wait. But there's a possibility, we don't meet tomorrow and so I otherwise I would just say let's revisit it tomorrow.
```
ljharb:
```suggestion
JHD: No, it's just a nice signal. I think stage one is a problem statement, right? So the specifics of the solution don't need to be explained or understood just yet but - like I'm happy to write up that explainer, it's more like, if the problem is unclear to someone such that they are not willing to or not comfortable, making it stage 1 without that explainer, then obviously it would need to wait. But there's a possibility we don't meet tomorrow, otherwise I would just say let's revisit it tomorrow. 
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+GC: I say given the confusion surrounding at least the discussion on this API in Matrix over the last ten minutes. think it would be good to have some. Some some things written down concretely that we can understand before discussing this more, 
+
+CP: yeah, you're on a weekend weekend, trailblaze these and get it next meeting for stage 1 and 2 maybe. I can't even get all the pieces in.
+
+WH: I'm also a bit unclear as to what is going into stage one and I would note that there is a deadline to post documents a number of days in advance of a meeting so we can review them.
+
+JHD:  Sure. And I mean that and the lack of materials are both valid, Reasons to procedurally, reject stage 1. I just but that doesn't mean you have to reject it but it sounds it sounds like they're not, there's at least a few people that aren't comfortable with stage one.
```
ljharb:
```suggestion
JHD:  Sure. And I mean that and the lack of materials are both valid reasons to procedurally reject stage 1 - but that doesn't mean you have to reject it. It sounds like there's at least a few people that aren't comfortable with stage 1 yet.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+WH: Yeah, but I also want to mention that I do like where you're going with this. 
+
+JWK: So I didn't catch up with the conversation before. What is the use case? Why do we need an API for it? 
+
+JHD: So that people can write code, that can't be trivially broken by someone later messing with the environment like node core uses this, because if do delete function.prototype.call node core just craps itself and dies immediately. And so there's all over nerdcore. They're trying to slowly patch this problem so that its robust against user modification of the platform and all of my npm packages are written in ways where as long as they run first. In a before user code is mess with anything, then they are robust against user code messing with things. And get intrinsic would be a much more convenient and potentially more performant although that's less of a concern for me way to to write these this robust code.
```
ljharb:
```suggestion
JHD: So that people can write code, that can't be trivially broken by someone later messing with the environment like node core uses this, because if I do `delete Function.prototype.call` node core just craps itself and dies immediately. All over node core, they're trying to slowly patch this problem so that its robust against user modification of the platform and all of my npm packages are written in ways where as long as they run first, before user code messes with anything, then they are robust against user code messing with things later. getIntrinsic would be a much more convenient and potentially more performant (although that's less of a concern for me) way to to write this robust code.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+ JWK:I guess that must be able to mock otherwise it might break visualization.
+
+KG: Right. So the assumption is that you are running first so that you can save off a copy of getOriginals that no one else will thereafter be in a position to modify.
+
+JWK: if they are running with why need you can just grab those by yourself
+
+JHD: I might Define a function first by the time that functions run a bunch of add code might have run or user extensions in their browser and mess with stuff and the in order for the the body of the function to be robust, I have to have cashed everything that it's going to use, and in to protect methods, I have to call bound to them in a way that they don't rely on.collared. Fine. Being present at runtime. And there's, I can already do this. I have a get intrinsic library that does this, but having it built into the platform, just like the Realms, proposal would make things a lot simpler and harder to get wrong.
```
ljharb:
```suggestion
JHD: I might define a function first, but by the time that function runs, a bunch of ad code might have run or user extensions in their browser, and messed with stuff, and in order for the the body of the function to be robust, I have to have cached everything that it's going to use, and in to protect methods, I have to call-bind them in a way that they don't rely on `.call` being present at runtime. I can already do this - I have a `get-intrinsic` library that does this, but having it built into the platform, just like the Realms proposal, would make things a lot simpler and harder to get wrong.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JWK: The realms use case seems much larger than getIntrinsics. It seems like a very limited use case. I don't know if it's worth it to do that.
+
+JHD: Yes, it's a fair point that there is a small number of people that may utilize that API, but I mean the that is a significant percentage of npm, download traffic already uses my get intrinsic Library so like it's the end users will will benefit from that. A lot of end users will benefit from the code even if they don't use it directly while. I think this may be true of realms.
```
ljharb:
```suggestion
JHD: Yes, it's a fair point that there is a small number of people that may utilize that API, but I mean the that is a significant percentage of npm's download traffic already uses my `get-intrinsic` library so the end users will benefit from that. A lot of end users will benefit from the code even if they don't use it directly. I think this may also be true of Realms.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JWK: One of the use cases I can imagine is that I can get the async function prototype without actually using an async function literal or that might not be able to ship to some old browsers.
+
+JHD: Could you rephrase that question? I'm not sure. I understood 
+
+JWK: some intrinsics require some newer syntax reach, for example, async function. AsyncGeneratorPrototype requires an async generator literal to get. And if we ship it, it breaks on the old browser.
+
+JHD: True, and my get intrinsic Library, currently handles that by using eval to evaluate. The syntax is such that it won't fail to parse an older browsers, but it will still work into her once, like, eval inside a try catch. It's gross. I would love to not have to ship that and, you know, be able to work in CSP environment, for example, but like the yeah, I mean so so then one day it is getting the native get intrinsic function is at this point already has all of the current syntax. So like that was alright, for a get intrinsic polyfill to solve which way. I
```
ljharb:
```suggestion
JHD: True, and my `get-intrinsic` library currently handles that by using `eval` to evaluate. The syntax is such that it won't fail to parse an older browsers, but it will still work with `eval` inside a try catch. It's gross. I would love to not have to ship that and, you know, be able to work in CSP environments, for example, but like the yeah, I mean so so then one day it is getting the native get intrinsic function is at this point already has all of the current syntax. So like that was alright, for a get intrinsic polyfill to solve which way.
```
i kind of trailed off fixing the end, but i think this still improves the beginning

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+JWK: Each room has its different solutions X. So X So, X must be the current realms 
+
+CZW: I mean I mean the article get coloring, the Gathering tricks and made some modifications on it like that adding some property on or either. My not change either in intrinsic. how it works. But it but it can be adding some next say, observable properties only. 
+
+JHD: Yeah. I mean if someone sticks like a foo property, an object that prototype and I the object, a prototype intrinsic, the object, I get will have a foo property. that is totally fine. I'm not trying to get an original snapshot. I'm just trying to be able to get the original functions primarily 
+
+CZW: So people can change the intrinsics when they getting run first and the later on the code cannot get the original intrinsics of a function. They will be seeing the modified one. If they are not...
+
+JHD: Yeah, you it's impossible and should remain, so to protect against code that runs before you, but it is possible. And should be much easier to protect against code that runs after you.
```
ljharb:
```suggestion
JHD: It is impossible and should remain so to protect against code that runs before you - but it is possible and should be much easier to do, to protect against code that runs after you.
```

meetings/2021-07/july-15.md
```diff
@@ -0,0 +1,734 @@
...
+
+USA: Do we have any consensus in this? No, Jordan. Do you think you'd like to come back to this next meeting? 
+JHD: Yeah, I think there's no consensus for stage 1 for it, but I haven't heard any obvious blockers for stage 1 except I need to have materials prepared before the deadline. So with that understanding, I will prepare it and consider it not a waste of my time and I will plan to come back at the next meeting with that request. So thank you, everyone. I'm sorry to hijack the realms discussion as much as I did.
```
ljharb:
```suggestion

JHD: Yeah, I think there's no consensus for stage 1 for it, but I haven't heard any obvious blockers for stage 1 except I need to have materials prepared before the deadline. So with that understanding, I will prepare it and consider it "not a waste of my time" and I will plan to come back at the next meeting with that request. So thank you, everyone. I'm sorry to hijack the realms discussion as much as I did.
```

ljharb (approved):
soooo sorry i missed the editing deadline and i have to spew all these github corrections into the PR :-(

PR #146, opened 2021-08-17
Add Phillip Mates (PMS)

ptomato:


robpalme (commented):
Would you consider `PMS` to follow the usual pattern?

philomates:
works for me; I updated the proposed change to PMS

gesa (approved):

PR #147, opened 2021-08-22
Add Aakash Patel (AVP)

avp:
I'm already in all the TC39 groups, but this never went through.


gesa (approved):
hey @avp, when were you on-boarded into the org? I just want to make sure we (the chair group) aren't missing any info we should have.

PR #153, opened 2021-09-16
Add notes for 2021-08 plenary

robpalme:
ref: https://github.com/tc39/Reflector/issues/388


ljharb (approved):

meetings/2021-08/sept-01.md
```diff
@@ -0,0 +1,585 @@
...
+
+JHD: And I’ll just say that if the, if implementations as a group are confident they can eventually ship it and they plan to and that they have no, you know, and as long as some implementations or implementations can ship it obviously because of the requirements for advancing stage for I’m comfortable with the personally, because once it’s part of the specification, I can build a polyfill and then that thing can just fall out of usage naturally over time. So I’m  thinking about, you know, the next ten years. Not the next you know, ten months. So it’s totally fine to me if there’s a delay, but thank you for sharing that concern.
+
+JWK: So, I have a question for the engine. Like, XS’s engine needs to be small. And what if, and I think that `getIntrinsic` is good. But it seems like we need to add too many strings into the engine, because those intrinsics are created by the strings like that. I guess that might add too much size to the embedded JS engines like XS. Maybe we can only add intrinsics that can only be reached by syntax. To get the list smaller.
```
Jack-Works:
```suggestion
JWK: So, I have a question for an engine like XS. They need to be small. I think `getIntrinsic` is good. But it seems like we need to add too many strings into the engine because those intrinsics are created by the strings like `%ArrayPrototype%.slice`. I guess that might add too much size to the XS. Maybe we can only add intrinsics that can only be reached by syntax to get the list smaller. 
```

meetings/2021-08/sept-01.md
```diff
@@ -0,0 +1,585 @@
...
+
+JWK: Well, so this makes me think of the polyfill. For example, we need to pull it through new intrinsics, and, in the current API design, we need to patch the whole `getIntrinsic` function. It’s possible to make it like a namespace, all intrinsics are the static properties on it. It’s like `Intrinsics.what`, so polyfills can add new intrinsics onto this namespace.
```
Jack-Works:
```suggestion
JWK: This makes me think of the polyfill. For example, we need to add new intrinsics, and in the current API design, we need to patch the whole `getIntrinsic` function. If it is like a namespace, all intrinsics are the static properties on it. E.g `Intrinsics["%ArrayPrototype%.slice"]`, so polyfills can add new intrinsics onto this namespace simply.
```

meetings/2021-08/sept-01.md
```diff
@@ -0,0 +1,585 @@
...
+
+JWK: Well, so this makes me think of the polyfill. For example, we need to pull it through new intrinsics, and, in the current API design, we need to patch the whole `getIntrinsic` function. It’s possible to make it like a namespace, all intrinsics are the static properties on it. It’s like `Intrinsics.what`, so polyfills can add new intrinsics onto this namespace. 
+
+JHD: I don’t think it is, because if that object is mutable, then it defeats the purpose of being robust. If that object is immutable then it prevents polyfills from adding it. So, like, literally the only option is a function that gives direct access that I’m aware of. We can still discuss that in Stage 1.
+
+JWK: You can run the polyfill first then save a copy of the Intrinsics object. Deep clone. So you get the polyfilled intrinsics and at once, to be modified later.
```
Jack-Works:
```suggestion
JWK: You can run the polyfill first (to add new intrinsics) then save a copy of the Intrinsics object with a deep clone. Then you get the polyfilled intrinsics at the initialization stage that won't be affected later.
```

meetings/2021-08/sept-01.md
```diff
@@ -0,0 +1,585 @@
...
+
+JHD: So I completely agree, but I think more users will be impacted by this than ever are affected by Atomics, for example, which is also a very niche use case, even on the Web Platform. I mean, I’m not, I’m not hearing to be hostile with that, but I just think that like, the amount of transitive code that depends on this pattern is very large.
+
+SYG: I think the point I’m trying to make is, I guess, the same point that I made earlier: that if a large part of this is so large, part of this is ergonomic. You want to cache one thing instead of _n_ things. I get that and I hear you there. The other cost is you don’t want to ship this heavyweight Library around, while you might not get around that cost. Anyway, even though you push it to the engine, and that is a thing, I would like a better handle on if it, in fact, has [?] effects on loading time. That might not be acceptable if it’s just a memory and we want to re-architect around that; maybe that’s okay. Thanks. 
+
+JHD: That’s very well understood. 
+
+JWK: Would want to come back. I suppose it is Stage 1. It’s worth the problem, too slow [?].
```
Jack-Works:
```suggestion
JWK: I suppose it to Stage 1. It’s a worth problem to solve. 
```

meetings/2021-08/sept-01.md
```diff
@@ -0,0 +1,585 @@
...
+
+SYG: I hear you and I intend to work closely with you. To the urgency question and the speed that I’m envisioning. This slide [slide 22, about strings] is I think the actual thing that might block implementations for a significant amount of time, and that is what I’m working on, and that is not blocked by, you know, standards progress. I think I want to get the ball rolling here. There are many interested parties and let’s try to nail down some design. That is amenable to everybody while this part which I think is the hard part is happening. 
+
+YSV: Yeah, for sure. For this, I’m going to have our GC folks take a look and work with you on that once they’ve got some cycles to do that.
+
+JWK: Currently on the web, the concurrent programming model is based on posting messages instead of memory sharing. So, adding a high-level abstraction of shared structs…that means we are encouraging the concurrent programming model based on memory sharing.
```
Jack-Works:
```suggestion
JWK: Currently on the web, the concurrent programming model is based on post messages instead of memory sharing. Adding a high-level abstraction of shared structs… That means we are encouraging the concurrent programming model based on memory sharing. 
```

meetings/2021-08/sept-01.md
```diff
@@ -0,0 +1,585 @@
...
+
+JWK: Okay, I think it’s fair too. Let’s have a JS that has [?] and supports multiple patterns.
```
Jack-Works:
```suggestion
JWK: Okay, I think it’s fair too. JS should be able to support multiple patterns (like FP and OOP).
```

meetings/2021-08/sept-01.md
```diff
@@ -0,0 +1,585 @@
...
+
+JWK: One of the primary use cases for shared memory and shared structs are for WebAssembly, but WebAssembly needs a shared struct because they need to handle the code that compiled from C++ or some other languages like that. So I think it’s acceptable to keep the shared struct inside multiple Wasm threads, but not let them into the JavaScript side. Multiple Wasm threads can [do?] programming by the shared data by the shared memory and if they want to send the results to JavaScript, they need to go through the message passing. I think that is better to have.
```
Jack-Works:
```suggestion
JWK: One of the primary use cases for shared memory and shared structs are for WebAssembly, WebAssembly needs a shared struct because they need to handle the code compiled from C++ or some other languages. I think it’s acceptable to keep the shared struct inside multiple Wasm threads, but not let them leaked into the JavaScript side. Multiple Wasm threads can program by the shared memory and if they want to send the results to JavaScript, they need to go through the message passing. I think that is better to have. 
```

meetings/2021-08/sept-01.md
```diff
@@ -0,0 +1,585 @@
...
+
+WH: Okay, then the problem is—the reason this works with SharedArrayBuffers is that when you get paradoxes caused by relaxed ordering, you just read or write some bad integer. On the other hand, with structs you don’t just get bad integers here and there when you get these hazards, you might get access to an object which has not been initialized yet. So this makes this unsafe, unless you make all accesses sequentially consistent.
+
+SYG: If that’s what needs to be done, then that is a direction. I had been thinking about it, but I thought unordered by default might be okay, but admit having not fully worked out the memory model.
+
+WH: You could make all accesses sequentially consistent, but then it would be too slow.
+
+SYG: Okay. Sorry, [where was?] I? Let’s continue this chat. This needs to be worked out. But not before Stage 1. I would like to ask for Stage 1 explicitly to explore the space of fixed layout objects for the following two use cases of shared memory and WasmGC. Any objections to Stage 1?
+
+JWK: I like the non-shared parts, but I think it’s skeptical for the shared parts, but I think Stage 1 is okay.
```
Jack-Works:
```suggestion
JWK: I like the non-shared parts, but the shared parts are skeptical. I think Stage 1 is okay though.
```

Jack-Works (changes requested):

PR #174, opened 2021-12-14
Add Mark Davis to delegates.txt as MED

rricard:


michaelficarra:
Mark Davis is Mark E Davis, who is already there.

PR #176, opened 2022-01-06
Transfer "super construct proposal" document from ecma262

ljharb:
Hi! This PR contains all these commits: https://github.com/tc39/ecma262/commits/83d11493e8c8bed337af1ccc58b53af0813c11cf/workingdocs/ES6-super-construct%3Dproposal.md which have been amended so the messages link to their original commit.

Once this lands, we can delete the document from ecma262 (since it doesn't really belong there).

I'd prefer to merge this with a merge commit, so that each commit remains in the git history on the default branch, which is important for archival purposes.


ryzokuken (approved):
Rubber stamp approval; I wasn't present at the meeting where this was discussed so I cannot attest to the contents here.

PR #204, opened 2022-07-09
Add notes for 2022-06

robpalme:
ref: https://github.com/tc39/Reflector/issues/430
ref: https://github.com/tc39/Reflector/issues/438


ljharb (approved):
(I don’t know if we need the “remote attendees” sections, that’s mostly for notetakers)

ryzokuken (approved):

PR #225, opened 2023-01-09
Added new delegate Ben Allen (BEN)

ben-allen:


delegates.txt
```diff
@@ -41,6 +41,7 @@ Augustus Yuan (AYN)
 Avik Chaudhuri (AVC)
 Axel Rauschmayer (ARR)
 Aysegul Yonet (AYS)
+Ben Allen (BEN)
```
ljharb: our typical convention would have this be `BAN`, is that ok?

ben-allen: Shoot, I knew snagging BEN was too good to be true. BAN is fine, though if possible BJA would be preferable

ljharb: Also, where's the matching reflector issue for this addition? Can you have your Ecma member company point of contact file one?

ryzokuken: @ljharb apologies, was waiting on the PR link before posting.

robpalme: Nudging this forwards...

```suggestion
Ben Allen (BAN)
```

ljharb (approved):

PR #233, opened 2023-02-27
Correct typo in reference to delegate

jugglinmike:
I was not present at this meeting, so @bterlson or @codehag should verify this change.


acutmore (approved):
Thanks!

PR #234, opened 2023-02-27
Fix misleading typo

jugglinmike:
I did not attend this meeting, so I'm not an authority on transcription errors. Judging from context, though, this seems to be exactly the opposite of what @ljharb was saying. Jordan: can you confirm?


ljharb (approved):
yep, this is the right change, thanks

PR #235, opened 2023-02-27
fix: Typescript -> TypeScript

johnnyreilly:
I'm a pedant, I'm so sorry.


acutmore (approved):
Much better! Thanks John

PR #238, opened 2023-03-20
Add LWT (Lenz Weber-Tronic) to delegates.txt

benjamn:
These are Lenz Weber-Tronic (@phryneas)'s initials, and they appear to be untaken so far in delegates.txt.

Lenz works with me at [Apollo Graph, Inc](https://apollographql.com), a current ECMA/TC39 member.


ljharb (approved):
Fixed the ordering.

PR #241, opened 2023-04-08
Add notes for March 2023

acutmore:
https://github.com/tc39/Reflector/issues/464


johnnyreilly:
I have found no mentions of `Typescript` - this is good

meetings/2023-03/mar-21.md
```diff
@@ -0,0 +1,1074 @@
...
... only be solved, very far down the lin when everybody has moved modules, and I don't think that that is a valid position or I don't think that that's a very strong position, and while I completely understand the concern about having a namespace that allows for possible collision having a mutable, object does not prevent you from using WeakMaps to resolve those potential collisions, It does prevent well-crafted and well, written code that can use a namespace like this. successfully from being able to do a very important feature in a script environment where they don't have access to modules to have that type of behavior. The other concern that we have is that – we've kind of rehash this over and over again on GitHub – but every solution that's been presented requires significant amounts of rewriting to support it, and we really don't think that that's valid or that's a viable option. So our strong position is that option? One does not prevent you from having the type of isolation you'd like out of option two but option two completely prevents the types of behaviors, we'd like to be able to use in a script environment. So we believe option, two is completely completely a. Non-starter
```
johnnyreilly: Should `decorato` be `decorator`?

```suggestion
RBN: So my concern. is and it's like KG has articulated, as being a TypeScript specific concern, but it really isn't my concern is not about library deduplication, although I do think that is a valid concern. My concern is about: Script is a valid parsing goal. there. I don't have numbers but my intuition is that the largest majority of existing JavaScript code run on the web is still scripts, not modules. A system that purely depends on the concept of public metadata being only available with modules is not actually public. And so the TypeScript specific example is that TypeScript would like to be able to at compile time as a compiler option, inject a decorator that adds metadata about the types of the things that you are decorating. This is the TypeScript emit decorator metadata flag, that we've had since 2015. and we'd like to be able to bring some of that capability forward and that is one of the main motivating reasons that we even introduced metadata as part of this proposal. For TypeScript to be able to work correctly and inject a decorator that can attach metadata in a script environments, I cannot depend on a module. depending on the module require an await of an import. Each. because I can't necessarily import. And I can't break the users expected. They expected semantics of code. that does not `await` currently by randomly injected in an await. And then, it could also be in the middle of an asynchronous function that you declare a class declaration. So that means that we cannot depend on modules. So if the only solution is to say that public metadata is only valid, if you're using modules. completely cuts off a very large percentage of people that are currently using JavaScript with scripts today in bundlers, etc., and saying that this solution can only be solved, very far down the lin when everybody has moved modules, and I don't think that that is a valid position or I don't think that that's a very strong position, and while I completely understand the concern about having a namespace that allows for possible collision having a mutable, object does not prevent you from using WeakMaps to resolve those potential collisions, It does prevent well-crafted and well, written code that can use a namespace like this. successfully from being able to do a very important feature in a script environment where they don't have access to modules to have that type of behavior. The other concern that we have is that – we've kind of rehash this over and over again on GitHub – but every solution that's been presented requires significant amounts of rewriting to support it, and we really don't think that that's valid or that's a viable option. So our strong position is that option? One does not prevent you from having the type of isolation you'd like out of option two but option two completely prevents the types of behaviors, we'd like to be able to use in a script environment. So we believe option two is completely completely a non-starter.
```

ctcpip (approved):

ljharb (approved):

PR #250, opened 2023-06-01
✅ add unit test for checkDelegates()

ctcpip:


ctcpip:
hide whitespace when viewing.  also switched cjs -> esm

package.json
```diff
@@ -1,7 +1,8 @@
 {
   "private": true,
+  "type": "module",
```
ljharb: let’s please use .mjs for esm instead of type module

scripts/check-delegates-test.js
```diff
@@ -0,0 +1,19 @@
...
+const twoLetter = `Chris de Almeida (CDA)\nRob Palmer (RP)\nUjjwal Sharma (USA)`;
+const threeLetter = `Chris de Almeida (CDA)\nRob Palmer (ROBPALMER)\nUjjwal Sharma (USA)`;
+const duplicate = `Chris de Almeida (CDA)\nRob Palmer (RPR)\nUjjwal Sharma (USA)\nUjjwal Sharma (USA)`;
+const valid = `Chris de Almeida (CDA)\nRob Palmer (RPR)\nUjjwal Sharma (USA)`;
+
+assert.throws(()=>checkDelegates(lex), { message: /Not in lexicographic order/});
+assert.throws(()=>checkDelegates(missing), { message: /Missing abbreviation for/});
+assert.throws(()=>checkDelegates(uppercaseLatin), { message: /Abbreviations must be all uppercase Latin letters/});
+assert.throws(()=>checkDelegates(twoLetter), { message: /not in allowlist. New delegate abbreviations must be three letters/});
+assert.throws(()=>checkDelegates(threeLetter), { message: /New delegate abbreviations must be three letters/});
```
ljharb:
```suggestion
assert.throws(() => checkDelegates(threeLetter), { message: /New delegate abbreviations must be three letters/ });
```
and similar throughout

ljharb (changes requested):

scripts/check-delegates-test.js
ctcpip: if we are enforcing styles, we should set it up via eslint/prettier

package.json
ctcpip: done

scripts/check-delegates-test.mjs
```diff
@@ -0,0 +1,19 @@
...
+const valid = `Chris de Almeida (CDA)\nRob Palmer (RPR)\nUjjwal Sharma (USA)`;
+
+assert.throws(()=>checkDelegates(lex), { message: /Not in lexicographic order/});
+assert.throws(()=>checkDelegates(missing), { message: /Missing abbreviation for/});
+assert.throws(()=>checkDelegates(uppercaseLatin), { message: /Abbreviations must be all uppercase Latin letters/});
+assert.throws(()=>checkDelegates(twoLetter), { message: /not in allowlist. New delegate abbreviations must be three letters/});
+assert.throws(()=>checkDelegates(threeLetter), { message: /New delegate abbreviations must be three letters/});
+assert.throws(()=>checkDelegates(duplicate), { message: /Conflicting usage on line/});
+
+assert.doesNotThrow(()=>checkDelegates(valid));
```
michaelficarra: You should also check for validity of a legacy two-letter abbreviation.

michaelficarra (commented):
You should check that the line numbers are being reported properly.

scripts/check-delegates-test.js
ljharb: i mean sure, but this is pretty universally idiomatic spacing

michaelficarra: @ctcpip That's not how this works lol. Just because we don't have CI style enforcement yet doesn't mean we have no style rules. Before widespread use of linters, it wasn't just a wild West.

Once we have a linter in CI, if someone asks you for further style changes that are not yet enforced by the linter, *then* you can respond "well the linter doesn't enforce it, so we should update the linter".

ctcpip: @michaelficarra Understood. For context, Jordan and I have collaborated on numerous PRs where we have style/lint changes and a recurring theme is that we don't have linting set up, so we go and set it up.

I don't disagree with the convention.

scripts/check-delegates-test.mjs
ctcpip: done

scripts/check-delegates-test.js
ctcpip: done

ctcpip: just noticed the white space reco for the curly -- fixed

.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  root: true,
+  extends: ["eslint:recommended"],
+  parserOptions: {
+    ecmaVersion: "latest",
+    sourceType: "module",
```
ljharb: this should be done in `overrides` only for `.mjs` files, not globally.

.nvmrc
```diff
@@ -1 +0,0 @@
-12
```
ljharb, line 1: this should probably be restored and set to `lts/*`

package.json
```diff
@@ -1,7 +1,12 @@
 {
   "private": true,
   "scripts": {
-    "check": "./scripts/check-delegates.js",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "check": "node -e 'import(\"./scripts/check-delegates.mjs\").then(cd => cd.checkDelegates())'",
+    "test": "node ./scripts/check-delegates-test.mjs",
+    "lint": "eslint . --ext .js --ext .mjs --ext .cjs",
+    "lint:fix": "eslint . --ext .js --ext .mjs --ext .cjs --fix"
```
ljharb:
```suggestion
    "lint": "eslint . --ext .js,.mjs,.cjs",
    "lint:fix": "npm run lint -- --fix"
```

mathiasbynens (approved):

.eslintrc.js
ctcpip: done

.nvmrc
ctcpip: done

ljharb (approved):

PR #254, opened 2023-06-08
💚 add markdownlint

ctcpip:


package.json
```diff
@@ -4,9 +4,13 @@
     "check": "node -e 'import(\"./scripts/check-delegates.mjs\").then(cd => cd.checkDelegates())'",
     "test": "node ./scripts/check-delegates-test.mjs",
     "lint": "eslint . --ext .js,.mjs,.cjs",
-    "lint:fix": "npm run lint -- --fix"
+    "lint:fix": "npm run lint -- --fix",
+    "//": "markdownlint commands ignore old files",
+    "mdlint": "markdownlint-cli2 '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
```
ljharb, line 9:
```suggestion
    "prelint": "npm run mdlint",
    "mdlint": "markdownlint-cli2 '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
```
then we don't need the line in the yml

.github/workflows/check.yml
```diff
@@ -14,5 +14,6 @@ jobs:
         with:
           node-version: lts/*
       - run: npm run lint
+      - run: npm run mdlint
```
ljharb:
```suggestion
```

package.json
ctcpip: @ljharb one less line in yml; one more line in json... I prefer the clarity and semantically-correctness of directly calling it

ljharb: the advantage is that someone locally only needs two commands to know, instead of three.

ctcpip: sure. my preference is still to leave it as is

ctcpip: I would favor a `lint:all` command if that's the motivation, but that's an additional line... 😱

ljharb: the idiomatic thing would be that `npm test` runs everything, meaning that `npm run lint` would be set up in `prelint`. iow, package.json being the source of truth is fine to have lots of config, because at least that centralizes it in the place everyone knows to look.

ctcpip: `prelint` appears to be an undocumented private API  🤔

ljharb (approved):

PR #265, opened 2023-08-08
Name change

pzuraq:
I've change my name recently


ctcpip (approved):

nicolo-ribaudo:
I think we should update all the CHG in the notes to KHG, so that they are still attributed to you and not to an undefined acronym. This is similar to what was done in https://github.com/tc39/notes/pull/55.

ctcpip:
@nicolo-ribaudo https://github.com/tc39/notes/pull/266

ctcpip:
as an aside, we probably should not have went forward with #55 ... LBR was the ideal TLA for Leo...  alas, here we are :)

PR #267, opened 2023-08-17
Add Pablo Gorostiaga Belio (PGB)

littledan:


ljharb (approved):

ctcpip:
`PBO` is the preferred convention.  can we use that instead?  @littledan

ljharb:
oh right, good call

ctcpip:
> `PBO` is the preferred convention.

resolved by https://github.com/tc39/notes/pull/274

ctcpip:
@gorosgobe Would you consider `PGO` or `PBO` to match others?

gorosgobe:
@ctcpip I don't mind either, but if we follow the convention of first and last letters of the last name, then it should be `PGO` - my last name is "Gorostiaga Belio"

ctcpip:
similar to me with `CDA` :)

PR #268, opened 2023-08-21
Add notes for July 2023

acutmore:


acutmore:
oh we have markdown linting now. yay! Will fix up and push

meetings/2023-07/july-11.md
```diff
@@ -0,0 +1,1683 @@
...
+| Willian Martins     | WMS          | Netflix           |
+| Sergey Rubanov      | SRV          | Invited Expert    |
+| Chris de Almeida    | CDA          | IBM               |
+| Michael Ficarra     | MF           | F5                |
+| Luca Casonato       | LCA          | Deno              |
+| Kevin Gibbons       | KG           | F5                |
+| Ron Buckton         | RBN          | Microsoft         |
+| Christian Ulbrich   | CHU          | Zalari            |
+| Tom Kopp            | TKP          | Zalari            |
+| Mikhail Barash      | MBH          | Univ. of Bergen   |
```
ljharb, line 40:
```suggestion
| Mikhail Barash      | MBH          | Univ. of Bergen   |
| Jordan Harband      | JHD          | Invited Expert    |
```

meetings/2023-07/july-11.md
```diff
@@ -0,0 +1,1683 @@
...
+
+CDA: To that end, we pruned inactive folks and then put out a call for additional participants. And to that end, we are thankful that both TAB and RCA have agreed to join us on the CoC Committee. The way that we handle new folks for the committee is a little different than we have done for the convenors, of the task groups, for example, we are not going to be doing that in real-time here. We ask if any objections to the new folks joining, again, that’s TAB and RCA, please let us know before the end of plenary. You can do that by either contacting, you know, the existing code of conduct email or just reaching out to one of the committee members directly. Those are currently myself, JHD and MM (?). That’s it for the CoC update. I see an item on the TCQ.
```
ljharb:
```suggestion
CDA: To that end, we pruned inactive folks and then put out a call for additional participants. And to that end, we are thankful that both TAB and RCA have agreed to join us on the CoC Committee. The way that we handle new folks for the committee is a little different than we have done for the convenors, of the task groups, for example, we are not going to be doing that in real-time here. We ask if any objections to the new folks joining, again, that’s TAB and RCA, please let us know before the end of plenary. You can do that by either contacting, you know, the existing code of conduct email or just reaching out to one of the committee members directly. Those are currently myself, JHD, and MPC. That’s it for the CoC update. I see an item on the TCQ.
```

meetings/2023-07/july-11.md
```diff
@@ -0,0 +1,1683 @@
...
+lot of decisions were made that we might not have made if we had expected that eventually we
+were going to have dynamic import and deferred imports and so on and so forth. So it’s sort of -- it just kind of feels like it leaves us in an awkward position, and I’m not sure if anyone still shares those original design goals and wants things to be maximally static or if
+we’ve just decided that’s not important anymore. And then related is there was an attempt to
+make a proposal for conditional static imports, and it kind of seems like this import deferral
+would be a way to do that, where you do an import defer and then you just conditionally access
+to property or not. So, yeah, I don’t know, I’m just -- I just kind of wanted to bring this
+all up. This is not a straightforward, obvious win for me. It’s -- like, the benefit that is
+being sought is valuable, but I’m just not -- I’m not super convinced on the limitations,
+fitting in the syntax and, you know, combining it with the mental model of ESM. I just wanted
+to bring that up.
```
ljharb:
```suggestion
JHD: I have a couple thoughts here. So one is that the -- like, I understand why you have to use `import *`, because it’s super weird and magic to have accessing a variable have effects. But one of the reasons -- like, `import *` is something that I’ve generally considered and found to be considered “gross”, for lack of a more precise word, sloppily and implicit and prevents a lot of static analysis. And, you know, it makes tree shaking more difficult things like that. And the original -- some of the original design goals of ESM were -- like, or a lot of them, seems to be “let’s make everything as static as possible”, so a lot of decisions were made that we might not have made if we had expected that eventually we were going to have dynamic import and deferred imports and so on and so forth. So it’s sort of -- it just kind of feels like it leaves us in an awkward position, and I’m not sure if anyone still shares those original design goals and wants things to be maximally static or if we’ve just decided that’s not important anymore. And then related is there was an attempt to make a proposal for conditional static imports, and it kind of seems like this import deferral would be a way to do that, where you do an import defer and then you just conditionally access to property or not. So, yeah, I don’t know, I’m just -- I just kind of wanted to bring this all up. This is not a straightforward, obvious win for me. It’s -- like, the benefit that is
being sought is valuable, but I’m just not -- I’m not super convinced on the limitations, fitting in the syntax and, you know, combining it with the mental model of ESM. I just wanted to bring that up.
```

ljharb (changes requested):
There's a ton of hard wraps here; is there any way we could remove them? for diff churn reasons as well as source editability, the notes should only have newlines between speakers.

acutmore:
> There's a ton of hard wraps here; is there any way we could remove them?

Yeah I think the transcript software adds them :(

I've had a go at removing them with a repeated find replace of `/(^[a-z][^\n]+)\n([a-z])/` -> `$1 $2` in vscode. And then fixing up the false positives. This seems to have got rid of most of them.

Would be nice if we could do this with markdownlint, maybe a custom rule if one doesn't already exist.

ljharb:
Even better is if the transcription software could avoid adding them in the first place. @littledan, is that something we could work with them on?

meetings/2023-07/july-11.md
```diff
@@ -0,0 +1,1227 @@
...
+
+KG: Just briefly. JHD, I agree this does make things a little more dynamic, but to me the most important property is that exports are not dynamic and that property continues to hold. As long as you don’t use computed properties to access things of on the name space object, it’s not like, less static in an important way as far as I’m concerned.
+
+JHD:  Yeah, I mean, I agree if you’re only using, like, dotted properties and you’re not passing  around the namespace object, then the dynamism is not really an issue.
```
ljharb, line 963:
```suggestion
JHD: Yeah, I mean, I agree if you’re only using, like, dotted properties and you’re not passing  around the namespace object, then the dynamism is not really an issue.
```
can we also remove more-than-one-space throughout? (but careful at the top where it'll mess up the attendees table to do a naive search and replace - i often do one for `.  ` → `. ` until there's nothing left, and `:  ` → `: `, etc, and then manually do the remaining `  `)

littledan:
@SaminaHusain is working on a version of the notes which includes additional summaries; let's work off of that PR (to be posted soon) instead.

littledan:
Re: the hard wraps, I will reiterate to our providers that they should improve their configuration to not add them. It continues to be that some captioners include them, and others don't.

ljharb:
We're already well past the normal 10 day timeline for posting notes; can the summaries be added on after landing this?

ctcpip:
> Would be nice if we could do this with markdownlint, maybe a custom rule if one doesn't already exist.

@acutmore: unfortunately markdownlint doesn't support multi-line fixes (and there is an output problem with custom rule for detection (https://github.com/OnkarRuikar/markdownlint-rule-search-replace/issues/110))

https://github.com/tc39/notes/pull/269 adds custom tooling for this, please take a look, especially at the regex.

PR #269, opened 2023-08-22
✨ add bad linebreaks linter+fixer

ctcpip:
tests would be nice, if someone wants to have a go  😃 .. otherwise I'll get to it eventually... probably...

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,18 +1,22 @@
 {
   "private": true,
   "scripts": {
+    "bad-linebreaks": "node ./scripts/bad-linebreaks.mjs",
+    "bad-linebreaks:fix": "npm run bad-linebreaks fix",
     "check-delegates": "node -e 'import(\"./scripts/check-delegates.mjs\").then(cd => cd.checkDelegates())'",
     "check-delegates-test": "node ./scripts/check-delegates-test.mjs",
+    "fix:all": "npm run lint:fix && npm run mdlint:fix && npm run bad-linebreaks:fix",
     "lint": "eslint . --ext .js,.mjs,.cjs",
     "lint:fix": "npm run lint -- --fix",
     "//": "markdownlint commands ignore old files",
     "mdlint": "markdownlint-cli2 '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
     "mdlint:fix": "markdownlint-cli2-fix '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
     "//": "markdownlint most likely to fail, so run that first",
-    "test": "npm run mdlint && npm run lint && npm run check-delegates-test && npm run check-delegates"
+    "test": "npm run mdlint && npm run bad-linebreaks && npm run lint && npm run check-delegates-test && npm run check-delegates"
   },
   "devDependencies": {
     "eslint": "^8.41.0",
+    "glob": "^10.3.3",
     "markdownlint-cli2": "^0.7.1"
   }
 }
--- a/scripts/bad-linebreaks.mjs
+++ b/scripts/bad-linebreaks.mjs
@@ -1,0 +1,72 @@
+#!/usr/bin/env node
+
+import fs from 'fs';
+import { glob } from 'glob';
+
+function getLine(txt, index) {
+
+  let line = 1;
+
+  for (let i = 0; i < index; i++) {
+    if (txt[i] === '\n') {
+      line += 1;
+    }
+  }
+
+  return line;
+
+}
+
+export function findBadLinebreaks(file, fix = false) {
+
+  let contents = fs.readFileSync(file, 'utf8').toString();
+
+  const re = /(?<=[\w\d ])\n(?=[\w\d])/g;
+  const matches = Array.from(contents.matchAll(re));
+
+  for (const m of matches) {
+
+    if (fix) {
+      contents = `${contents.slice(0, m.index).trimEnd()} ${contents.slice(m.index + 1)}`;
+    } else {
+
+      const start = Math.max(0, m.index - 33);
+      const end = Math.min(contents.length - 1, m.index + 33);
+
+      console.log(`found erroneous linebreak at line ${getLine(contents, m.index)}:\n${contents.slice(start, end)}\n`);
+
+    }
+
+  }
+
+  if (matches.length > 0) {
+
+    if (fix) {
+      fs.writeFileSync(file, contents);
+      console.log(`fixed ${matches.length} erroneous linebreaks`);
+    } else {
+      process.exitCode = 1;
+    }
+
+  }
+
+}
+
+// patterns match what is in package.json mdlint scripts
+const files = await glob(
+  '**/*.md',
+  {
+    ignore: [
+      'node_modules/**',
+      'meetings/201*/*.md',
+      'meetings/202[0-2]*/*.md',
+      'meetings/2023-0[1-3]/*.md',
+    ]
+  }
+);
+
+const fix = process.argv?.[2] === 'fix';
+
+for (const f of files) {
+  findBadLinebreaks(f, fix);
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
   },
   rules: {
     "arrow-spacing": "error",
+    "brace-style": ["error", "1tbs"],
     "object-curly-spacing": ["error", "always"]
   },
   overrides: [


package.json
```diff
@@ -1,6 +1,8 @@
 {
   "private": true,
   "scripts": {
+    "bad-linebreaks": "node ./scripts/bad-linebreaks.mjs",
+    "bad-linebreaks:fix": "node ./scripts/bad-linebreaks.mjs fix",
```
ljharb: this isn't needed, since you can `npm run bad-linebreaks fix`

scripts/bad-linebreaks.mjs
```diff
@@ -0,0 +1,74 @@
...
+
+  const re = /(?<=[\w\d ])\n(?=[\w\d])/g;
+  const matches = Array.from(contents.matchAll(re));
+
+  for (const m of matches) {
+
+    if (fix) {
+      contents = `${contents.substring(0, m.index)} ${contents.substring(m.index + 1)}`;
+    }
+    else {
```
ljharb:
```suggestion
    } else {
```
elses need cuddles

scripts/bad-linebreaks.mjs
```diff
@@ -0,0 +1,74 @@
...
+    ignore: [
+      'node_modules/**',
+      'meetings/201*/*.md',
+      'meetings/202[0-2]*/*.md',
+      'meetings/2023-0[1-3]/*.md',
+    ]
+  }
+);
+
+const fix = process.argv[2] ? process.argv[2] === 'fix' : false;
```
ljharb:
```suggestion
const fix = process.argv?.[2] === 'fix';
```

scripts/bad-linebreaks.mjs
```diff
@@ -0,0 +1,74 @@
...
+  let line = 1;
+
+  for (let i = 0; i < index; i++) {
+    if (txt[i] === '\n') {
+      line += 1;
+    }
+  }
+
+  return line;
+
```
ljharb, line 17:
```suggestion
    return txt.slice(0, index).replace(/[^\n]/g, '').length + 1;
```
would this work?

scripts/bad-linebreaks.mjs
```diff
@@ -0,0 +1,74 @@
...
+  let contents = fs.readFileSync(file, 'utf8').toString();
+
+  const re = /(?<=[\w\d ])\n(?=[\w\d])/g;
+  const matches = Array.from(contents.matchAll(re));
+
+  for (const m of matches) {
+
+    if (fix) {
+      contents = `${contents.substring(0, m.index)} ${contents.substring(m.index + 1)}`;
```
ljharb:
```suggestion
      contents = `${contents.slice(0, m.index)} ${contents.slice(m.index + 1)}`;
```
lol, this would actually be nice to have `contents.with(m.index, ' ')` for

scripts/bad-linebreaks.mjs
```diff
@@ -0,0 +1,74 @@
...
+    }
+    else {
+
+      const start = Math.max(0, m.index - 33);
+      const end = Math.min(contents.length - 1, m.index + 33);
+
+      console.log(`found erroneous linebreak at line ${getLine(contents, m.index)}:\n${contents.substring(start, end)}\n`);
```
ljharb:
```suggestion
      console.log(`found erroneous linebreak at line ${getLine(contents, m.index)}:\n${contents.slice(start, end)}\n`);
```

ljharb (commented):
Looks good :-) lots of empty line padding tho, makes it a bit hard to read.

package.json
OnkarRuikar: > this isn't needed, since you can `npm run bad-linebreaks fix`

This is self documenting the availability and usage of the `fix` option. Without it a contributor will have to look at bad-linebreaks.mjs script.

package.json
```diff
@@ -9,10 +11,11 @@
     "mdlint": "markdownlint-cli2 '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
     "mdlint:fix": "markdownlint-cli2-fix '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
     "//": "markdownlint most likely to fail, so run that first",
-    "test": "npm run mdlint && npm run lint && npm run check-delegates-test && npm run check-delegates"
+    "test": "npm run mdlint && npm run bad-linebreaks && npm run lint && npm run check-delegates-test && npm run check-delegates"
```
OnkarRuikar, line 15:
```suggestion
    "test": "npm run mdlint && npm run bad-linebreaks && npm run lint && npm run check-delegates-test && npm run check-delegates",
    "fix:all": "npm run bad-linebreaks:fix && npm run lint:fix && npm run mdlint:fix"
```
We can create a `CONTRIBUTING.md` doc and [mention to run `npm run fix:all` before making the commit](https://github.com/OnkarRuikar/markdownlint-rule-search-replace/blob/main/CONTRIBUTING.md#dev-tooling).

scripts/bad-linebreaks.mjs
OnkarRuikar: In case of `abc \nefg` this will create `abc  efg`.
```suggestion
      contents = `${contents.slice(0, m.index)}.trimEnd() ${contents.slice(m.index + 1)}`;
```

scripts/bad-linebreaks.mjs
OnkarRuikar: > would this work?

Aren't string operations costlier than simple traversal?

package.json
ljharb: if it was turned into `--fix` it'd match eslint and there'd be no need to self-document it.

scripts/bad-linebreaks.mjs
ljharb: i'm not sure why performance would literally ever matter here

package.json
```diff
@@ -9,10 +11,11 @@
     "mdlint": "markdownlint-cli2 '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
     "mdlint:fix": "markdownlint-cli2-fix '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
```
OnkarRuikar, line 13: The `markdownlint-cli2-fix` [is deprecated](https://github.com/DavidAnson/markdownlint-cli2#command-line). Use `--fix` option instead.
```suggestion
    "mdlint:fix": "markdownlint-cli2 --fix '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
```

acutmore:
Thanks @ctcpip! When I was manually regex fixing this I noticed that I got false positive match on code blocks (I had copied some code that the speaker had pasted into Matrix, so the notes would have the relevant context).

Maybe we can detect "```" blocks and ignore them. Long term feels like we should use a proper markdown parser in case there are other cases?

OnkarRuikar:
> Maybe we can detect "```" blocks and ignore them. Long term feels like we should use a proper markdown parser in case there are other cases?

Just a suggestion. May be following could work:\
The markdownlint plugin [markdownlint-rule-search-replace](https://github.com/OnkarRuikar/markdownlint-rule-search-replace#using-markdownlintjson-config-file) has capability to narrow scopes to only code or only text. [[example code](https://github.com/mdn/content/blob/d546a54139681cad9b176db2ca494b6a6bb8f7c6/.markdownlint-cli2.jsonc#L132)]

Config provided in [this discussion](https://github.com/OnkarRuikar/markdownlint-rule-search-replace/issues/110#issuecomment-1689296404) will work to flag the hard wrap cases. Auto fixing, for this case, would require a separate script which could get the locations from the logs.

package.json
ctcpip: it's not needed, but strong preference for leaving it.  this is the place where people will look for this when CI fails, and `lint:fix` is a common pattern

scripts/bad-linebreaks.mjs
ctcpip: this is a lot slower, which is why I favored the unga bunga loop

package.json
ctcpip: that's new -- not available in this version.  we can upgrade but let's do that in a separate PR

ctcpip:
> Looks good :-) lots of empty line padding tho, makes it a bit hard to read.

sir, please.  these are my emotional support linebreaks

scripts/bad-linebreaks.mjs
ctcpip: for you?  sure

package.json
ctcpip: sure.  adding something to the readme would be sufficient

scripts/bad-linebreaks.mjs
ctcpip: I was thinking this would get run after mdlint fixes, which would trim the ends, but sure, why not

ctcpip:
@acutmore 

> When I was manually regex fixing this I noticed that I got false positive match on code blocks

ah yes...

````
```javascript
a = { b: x };
a = ({ b: x });
( { b: x } = a );
( ({ b: x }) = a ); // does not work
```
````

right now, this code will collapse the first two lines.  so in this case, I think the fix to appease this linter will be:

````
```javascript

a = { b: x };
a = ({ b: x });
( { b: x } = a );
( ({ b: x }) = a ); // does not work
```
````

not ideal, but an acceptable trade-off.  agree implementing this with a proper lexer would be a lot better

**Comment by ctcpip** [comment]
I've addressed all comments that I'm willing to address 😄

I'll follow on with a PR to fix the linting errors

File: scripts/bad-linebreaks.mjs (hunk 4 above)
Comment (ljharb) [reply scripts/bad-linebreaks.mjs:17]: in theory sure, but how impactful is it in practice?

File: package.json (hunk 1 above)
Comment (ljharb) [reply package.json]:
```suggestion
    "bad-linebreaks:fix": "npm run bad-linebreaks fix",
```
then let's at least do this so we're not repeating the command

ctcpip: sure? 🤷

scripts/bad-linebreaks.mjs
ctcpip: I agree that perf here is not paramount.  I still prefer what's there over the regex alternative.

package.json
```diff
@@ -1,18 +1,22 @@
...
     "check-delegates-test": "node ./scripts/check-delegates-test.mjs",
+    "fix:all": "npm run lint:fix && npm run mdlint:fix && npm run bad-linebreaks:fix",
     "lint": "eslint . --ext .js,.mjs,.cjs",
     "lint:fix": "npm run lint -- --fix",
     "//": "markdownlint commands ignore old files",
     "mdlint": "markdownlint-cli2 '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
     "mdlint:fix": "markdownlint-cli2-fix '**/*.md' '!node_modules' '!meetings/201*/*.md' '!meetings/202[0-2]*/*.md' '!meetings/2023-0[1-3]/*.md'",
     "//": "markdownlint most likely to fail, so run that first",
-    "test": "npm run mdlint && npm run lint && npm run check-delegates-test && npm run check-delegates"
+    "test": "npm run mdlint && npm run bad-linebreaks && npm run lint && npm run check-delegates-test && npm run check-delegates"
```
ljharb, line 15:
```suggestion
    "pretest": "npm run mdlint && npm run bad-linebreaks && npm run lint",
    "test": "npm run check-delegates-test && npm run check-delegates"
```

scripts/bad-linebreaks.mjs
ljharb: i was serious about this one, it's the overwhelmingly idiomatic style for JS

ljharb (approved):
also was serious about the excessive blank line paddings :-p

scripts/bad-linebreaks.mjs
ctcpip: weird hill to die on

ctcpip: you get brace style, I keep my emotional support whitespace

package.json
ctcpip: sir...

"my ideal is that eventually npm test is sufficient to run every CI check" -Jordan Harband

ljharb: yes, and `npm test` automatically runs `npm run pretest` before it and `npm run posttest` after it (that's how all npm scripts work)

ctcpip: ah that's right.. I forget this regularly...  but I'm also a fan of being explicit about what's happening and not relying on hidden magic (even if common knowledge)

scripts/bad-linebreaks.mjs
ctcpip: @OnkarRuikar just FYI, had to remove the `trimEnd()` because it throws off the indexing:

e.g.:

```
CCC: This 
is 
a 
sentence with trailing spaces. 
```

becomes

```
CCC: This is  
se tence with trailing spaces. 
```

of course can just run the regex again against the new string in the loop, but I'm not going to bother for now.  shouldn't be an issue though since we should be running the package.json scripts which will clean up the line endings before this runs.

PR #271, opened 2023-08-25
✨ use lexer for md processing, add test

ctcpip:
## tl;dr: this is an improvement over the existing impl and reduces false positives.  output from input of recent notes looks good

tried to fix more of the md problems, but as expected, end up fixing some things and then creating new problems... always going to be a trade-off..  hoping this will be sufficient.  ~draft status until I do more testing against notes files~ test results LGTM

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,6 @@ indent_size = 2
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[scripts/test-samples/*]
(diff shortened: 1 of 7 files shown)


ctcpip:
result of running the fix on the last meeting notes here:  https://github.com/tc39/notes/pull/273

ljharb (approved):
<br />
<br />
<br />
LGTM
<br />
<br />
<br />
;

ctcpip:
you're an anti-scrollite!

PR #274, opened 2023-08-25
✅ add test for ideal TLAs

ctcpip:


scripts/check-delegates.mjs
```diff
@@ -19,8 +19,44 @@ export function checkDelegates(contents = fs.readFileSync('./delegates.txt', 'ut
     'VM', 'WH', 'YK', 'ZB',
   ]);
 
+  // list of abbreviations that are not ideal.
+  // most of these are here because they are grandfathered in.
+  // new elements should only be false positives.
+  const NON_IDEAL_ABBRS = new Set([
```
michaelficarra, line 34: How was this list generated? I would think it would either have no 2-letter abbreviations (preferred) or all 2-letter abbreviations. But instead it only has some? How do we know that each entry is necessary?

michaelficarra (commented):
Is this really necessary? I don't think we need to be failing CI over this. And I expect it will have a pretty high false-positive rate. Overall, doesn't seem worth it.

scripts/check-delegates.mjs
ctcpip: it was generated by replacing line 100 with pushing to an array to write out all the existing 'non-ideal' entries that appeared

michaelficarra: Then how do you not have an entry for all 2-letter abbreviations?

ptomato:
We are a global organization so it seems less ideal to codify rules depending on the existence of a single first name and single last name - see [Falsehoods Programmers Believe About Names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)

ctcpip:
> Is this really necessary?

_necessary_ is a strong word, but this has been a recurring problem with folks creating TLA PRs and it takes the chairs an inordinate amount of time and extended feedback loops to resolve.  the point of this PR is to improve the process, catch problems and reduce the burden overall.

an alternative could be to just move the TLA PR onboarding item into the 'chair tasks' portion of onboarding -- and it's something we often already do anyway.  edit: but chairs can, and do, still make mistakes, so these checks are still useful.

edit:

> We are a global organization...

agree -- I hope my above comments clear this up.  this isn't about drawing lines in the sand, only about trying to better adhere to our convention and avoid mistakes.

scripts/check-delegates.mjs
ctcpip: last week seems like an eternity ago..  it was running on an earlier version of this code, and IIRC, I had originally combined the 2 letter check with the ideal check, but this resulted in messing with determinism and the test assertions, so I separated it out and moved the check to be last.  will update, good catch.

ptomato:
Got it, I understand the motivation. In that case, how important is it that this particular convention be followed? As long as people put their abbreviations in their video chat names, and on their nameplates when in person... can't we just allow any TLA that is reasonably associated with the delegate's full name, and cut out the extra work for the chairs that way?

scripts/check-delegates.mjs
ctcpip: fixed

ctcpip:
> how important is it that this particular convention be followed?

this is a fair question and one that I can't answer definitively.  one of the biggest reasons for the convention is so that when people taking notes are doing attribution, if they don't have all the TLAs memorized, they can reasonably infer based on the convention.  e.g. if they see Rob Palmer is talking, that will infer `RPR`, which is correct in that case.  of course, it's imperfect and there are numerous exceptions.

to be clear, there is nothing with this PR that says we can't add new exceptions to the convention; it's only to make it so that this is done deliberately and not accidentally.  this PR doesn't change the convention nor does it change anything about following or not following the convention.

delegates.txt
```diff
@@ -336,7 +336,7 @@ Nikolaus Papaspyrou (NPU)
 Noah Tye (NTE)
 Norbert Lindenberg (NL)
 Oliver Hunt (OH)
-Pablo Gorostiaga Belio (PGB)
+Pablo Gorostiaga Belio (PBO)
```
littledan: Why override the name selected by the delegate?

littledan:
If we want to reform acronyms of people, I'd rather focus on moving people with two-letter acronyms to three-letter ones, so we aren't encoding this irrelevant seniority information. This patch goes in the opposite direction, making an *additional* marker of semi-seniority for people with irregular TLAs. I prefer our system of letting people choose whatever TLA they want.

delegates.txt
ctcpip: I asked if it was ok to follow the typical convention -- https://github.com/tc39/notes/pull/267#issuecomment-1682811818

ctcpip:
> If we want to reform acronyms of people

that's not what we're doing here

> an additional marker of semi-seniority for people with irregular TLAs

that's not the intent here, please [see this comment](https://github.com/tc39/notes/pull/274#issuecomment-1696148851)

> I prefer our system of letting people choose whatever TLA they want.

that's not entirely accurate -- we steer toward the convention if available

delegates.txt
littledan: Sorry, I missed that and didn't pass it on to the delegate. Did you get in touch with him?

ctcpip: I did not, but will.  This PR does not need to be gated on that at all, I just needed to either make the change to `delegates.txt` or to the test code because his TLA PR already merged.

nicolo-ribaudo:
> this is a fair question and one that I can't answer definitively. one of the biggest reasons for the convention is so that when people taking notes are doing attribution, if they don't have all the TLAs memorized, they can reasonably infer based on the convention. e.g. if they see Rob Palmer is talking, that will infer RPR, which is correct in that case. of course, it's imperfect and there are numerous exceptions.

If this is the problem we are trying to solve, can we instead force people (both remote, in the zoom name, and in person, on a piece of paper), to explicitly write their acronym? Even to the point of only letting people speak if they have the acronym spelled out.

This would be incredibly more helpful than asking note takers to infer the acronym based on the name, since note taking is already a demanding task and we should try to reduce the cognitive overload, not to increase it.

ctcpip:
It's my fault for not including an explainer when I opened this PR, but the intent of this has been misunderstood and my attempts at clarification seem to have been inadequate.

This PR is not about changing anything, reforming anything, or doing anything different from the status quo.  It is simply to call attention to _potential_ issues to reduce the incidence of human error -- which is the same point of the existing tests.  We can assess the merits of the test and my thinking is that, if this is merged and then becomes clear it is not worth having, then we can simply remove it.

Issues regarding the TLA convention, improving outcomes for note-taking during meetings, and other aspects that have been brought up are all valid and are fair game for discussion, but are beyond the scope of this PR.  Those topics would be better served via issues in Reflector.

PR #282, opened 2023-09-06
✨ fix extra whitespace problems in bad-linebreaks run

ctcpip:
see https://github.com/tc39/notes/pull/283 for the fix PR for status checks

I didn't rename `bad-linebreaks.mjs` (and corresponding script name) to be less specific now that its scope is more than linebreaks, but we can change


ctcpip:
I didn't rename `bad-linebreaks.mjs` (and corresponding script name) to be less specific now that its scope is more than linebreaks, but we can change

ryzokuken (approved):

ljharb (approved):

PR #285, opened 2023-09-08
Update delegates.txt - Arpad Borsos

ctcpip:


ctcpip:
@Swatinem the conventional acronym for you is `ABS`, but that is already taken.  would you be ok with `ARS` or would you prefer something else?

Swatinem:
That sounds okay, can’t think of anything else :-)

ljharb (approved):

PR #296, opened 2023-10-24
Add Eric Leese (ESL)

syg:


ctcpip:
@EricSL Would you consider `ELE` to match others?

syg:
> @EricSL Would you consider ELE to match others?

What's wrong with ESL?

ctcpip:
when available, conventionally the TLA should be first letter of first name, first letter of last name, last letter of last name.

syg:
I am not aware that we had a convention we tried to follow. I am also not in favor of following a convention. We ask delegates for their preference, and if there's no conflict, we should honor their preference.

ctcpip:
it's not a hard requirement, just where we steer when available.  nothing wrong with ESL

ctcpip (approved):

PR #297, opened 2023-10-26
September 2023

SaminaHusain:
Please see the technical notes from the Tokyo meeting.

PR #298, opened 2023-10-31
✅ add additional example to test assertions

ctcpip:
don't merge this yet...


acutmore (approved):

scripts/test-samples/bad-linebreaks-sample-before.md
```diff
@@ -59,6 +59,21 @@ let biscuits = "delicious";
   *     one   more       
   +     time  
 
+RPR: thanks especially to Roderigo for helping out to arrange that meeting. We can see you 
+in the room, Roderigo. Thank you. All right. So on to the usual items. So let’s start, 
+let’s see if we have approval for last -- actually, I don’t think we fully committed them yet, 
+so we should probably -- let’s return to that perhaps later in the meeting once we’ve got that.
```
acutmore, line 65: This isn't quite the same, it has trailing whitespace.

```suggestion
so we should probably -- let’s return to that perhaps later in the meeting once we’ve got that.
```

ctcpip: yes, the trailing whitespace is in the original notes!  but this is also the key difference in why this succeeds vs after the markdownlint fix.  will elaborate in the other PR.  edit:  see https://github.com/tc39/notes/pull/297#discussion_r1378197564

ctcpip:
@acutmore  alright, I think we're good to go.  have a look at the new regex, before/after test examples, and if you like, the resulting changes to existing notes.  @ljharb FYI I'd prefer to not squash these commits.

PR #318, opened 2024-04-08
Add Jatin Ramanathan (JRA)

syg:


michaelficarra (approved):

ctcpip:
@michaelficarra not a big deal, but in the future please note this is part of the [chair tasks](https://github.com/tc39/Admin-and-Business/issues/431#issue-2231342910) for onboarding new delegates and we do this in a coordinated manner

michaelficarra:
@ctcpip Apologies. I knew it was a chair responsibility, but not that it was meant to be exclusive to chairs. I was concerned about this delegate being attributed in the notes without officially being in `delegates.txt` so I acted hastily.

ctcpip:
@michaelficarra - no worries. the biggest reason for it is we have had people merging prematurely in the past. that did not happen here though.  this situation (where attribution is needed but TLA not merged yet) usually doesn't happen because we make sure to complete outstanding onboarding for folks prior to plenary. in this case, onboarding issues were created during plenary

PR #343, opened 2024-09-06
Add Dominic Farolino (DMF)

syg:


delegates.txt
```diff
@@ -128,6 +128,7 @@ Dmitry Lomov (DL)
 Dmitry Makhnev (DJM)
 Dmitry Soshnikov (DS)
 Domenic Denicola (DD)
+Dominic Farolino (DMF)
```
robpalme, line 131: Is there any reason not to use the regular pattern?

```suggestion
Dominic Farolino (DFO)
```

robpalme (approved):

delegates.txt
syg: Because I asked him and he preferred `DMF` over `DFO`. This is the second time I've gotten pushback on the TLA that the delegate prefers. If chairs would prefer to always just use the regular pattern, please remove the sentence from the template that says to work with the delegate to find their preferred TLA.

robpalme: This is checking - not pushback - as part of a positive approval.  I'll see if we can clarify the template to make this slicker.

PR #363, opened 2025-03-07
February, 2025 notes

gesa:
One could argue this PR closes tc39/Reflector#547.

### DIFF
--- a/meetings/2025-02/february-18.md
+++ b/meetings/2025-02/february-18.md
@@ -1,0 +1,1554 @@
+# 106th TC39 Meeting | 18 February 2025
+
+**Attendees:**
+
+| Name             | Abbreviation | Organization       |
+|------------------|--------------|--------------------|
+| Kevin Gibbons    | KG           | F5                 |
+| Keith Miller     | KM           | Apple Inc          |
+| Oliver Medhurst  | OMT          | Invited Expert     |
+| Dmitry Makhnev   | DJM          | JetBrains          |
+| Gus Caplan       | GCL          | Deno Land Inc      |
+| Daniel Ehrenberg | DE           | Bloomberg          |
+| Jesse Alama      | JMN          | Igalia             |
+| Michael Saboff   | MLS          | Apple Inc          |
+| Ujjwal Sharma    | USA          | Igalia             |
+| Ashley Claymore  | ACE          | Bloomberg          |
+| Nicolò Ribaudo   | NRO          | Igalia             |
+| Philip Chimento  | PFC          | Igalia             |
+| Michael Ficarra  | MF           | F5                 |
+| Linus Groh       | LGH          | Bloomberg          |
+| Samina Husain    | SHN          | Ecma               |
+| Ron Buckton      | RBN          | Microsoft          |
+| Kris Kowal       | KKL          | Agoric             |
+| Mikhail Barash   | MBH          | Univ. of Bergen    |
+| Daniel Minor     | DLM          | Mozilla            |
+| Aki Rose Braun   | AKI          | Ecma International |
+| Luis Pardo       | LFP          | Microsoft          |
+| Chip Morningstar | CM           | Consensys          |
+| Eemeli Aro       | EAO          | Mozilla            |
+| Ben Lickly       | BLY          | Google             |
+| Mathieu Hofman   | MAH          | Agoric             |
+| Sergey Rubanov   | SRV          | Invited Expert     |
+| Chris de Almeida | CDA          | IBM                |
+| Luca Casonato    | LCA          | Deno               |
+| Istvan Sebestyen | IS           | Ecma International |
+| Waldemar Horwat  | WH           | Invited Expert     |
+| Richard Gibson   | RGN          | Agoric             |
+| Shane F Carr     | SFC          | Google             |
+| Erik Marks       | REK          | Consensys          |
+| Justin Grant     | JGT          | Invited Expert     |
+
+## Opening & Welcome
+
+Presenter: Rob Palmer (RPR)
+
+RPR: Thanks everyone for coming to Seattle, we’re ready to begin the 106th meeting at TC39.
+
+## Secretariat comments
+
+Presenter: Samina Husain (SHN)
+
+* [slides](https://github.com/tc39/agendas/blob/main/2025/tc39-2025-005.pdf)
+
+SHN: I was not able to attend your in-person meeting in Tokyo. It’s nice to see everybody again. A lot has happened within Ecma in the last couple of months. So let me just give you a short report. This is the overview of what we have done within the secretariat and different activities and just going to highlight some things on source map and new TCs we have and just a reminder on the executive committee meeting date and deadlines that you should all be aware of and then some general collaborations that we continue to have.
+
+SHN: So first congratulations for the first addition of source maps. Yes, there you are NRO. Congratulations. Good work. Really hard work. I mean, it was brilliant that you did this in the year. So we’re looking forward to the next edition. It is published and was approved at the December GA and want to bring that to everybody’s attention if you were not already aware.
+
+RPR: Round of applause.
+
+SHN: Also want to highlight TC55, a lot of work has gone into that. I want to thank everybody for that work. The chairs are Luca and Andreu, and Aki is supporting you and I think you will have the official meetings. I know there’s some things that need to be taken care of administratively from Ecma perspective. This is brilliant. Last year 2024 was the year that we had two new TCs and this is one of them. We’re very happy for that. That was approved at the December GA. If there are individuals that want to be involved and organizations that would be great.
+
+SHN: The other TC that is TC56 is the first that touches on artificial intelligence, I think this is very good for ECMA to come into a new space. The members are involved here not only IBM but ServiceNow has shown interest and Microsoft shown interest and Hitachi shown interest that are current members and university that are active not for profit members and we have a couple of invited experts and looking to have them transition to be membered and mainly Cisco being one of them. I think that is excellent. This also enables us to be more visible in the AI space. The AI Alliance has a number of individuals involved in the meeting. Hopefully this grows not as this particular TC but other TCs in the area of AI which of course as everyone knows is a dynamic and constantly changing topic. So thank you for the committee that worked on this also. It was approved in December.
+
+SHN: Also from the Ecma new management for 2025, we had the final votes in December and first I want to thank all of the people who did nominate and were nominated for this. We had seven members who are nominated for non-ordinary member positions. We had only four positions. It’s excellent to see the interest. I hope your interest still remains and that you are still active and want to consider to submit your name as we do these elections every year. This is the final slate that was selected. So DE who is sitting right here in the room is our new President of Ecma. And Jochen Friedrich was previous president and vice president and treasurer is Luoming Zhang and important to note that management roles are for ordinary members. We only have six ordinary members. We are looking forward to their support and driving Ecma.
+
+SHN: From the Executive Committee (ExeCom), this is really exciting and we have new members and many are sitting here. Jochen will chair and we have Theresa O’Connor from Apple and Chris Wilson from Google and Mikhail Barash (University of Bergen) and Patrick Dwyer (ServiceNow) and I don’t know if some are online. It’s great that we have new thoughts and new discussions. Peter Hoddie (Moddable) has been on it before and also important to remember all of the relevance and the past information and leverage on that. So it’s great that Peter is on. And Ross Kirsling (Sony) is also on it. Maybe Ross is also online. This is very new and dynamic executive committee. We will have our first meeting coming up in March. So again thank you everybody for your interest. And congratulations to everyone that was appointed. You may applaud since Daniel was here.
+
+SHN: Just a few general items and the important one, this year ExeCom is earlier usually it is middle or end of April. It is in March. If you are, as I assume, going to have addition 16 and 12, the different standards, I need to have an indication of that. We need to bring that to the attention of the ExeCom that is actually in two weeks. I can make that update immediately. But I just want to get information from the committee that both those two additions are your intentions for the GA in June. You have plenty of time for the opt-out, but I do need indication that you want it. And then maybe the editors can let me know that. The committee can let me know that before this meeting is over. Perhaps if you can by the end of the day, I would appreciate that.
+
+SHN: We have talked about liaisons before. Ecma has a number of liaisons that we keep active like the JTC 1 and W3C and we have had historical with IETF and in the past there were people of IETF coming to Ecma and TC39 meetings. Over time, there hasn’t been a lot of cross-contribution. I would not want to see the liaison disappear. To build a new liaison is always more complicated. I would like to keep it going. I’m asking if there’s somebody on the committee that could be the point of contact between TC39 and IETF, I would appreciate that. If there is no topic or nobody of interest, I also need to know. I need to figure out how to maintain that from the secretariat. I don’t want to drop this. I suspect as we move forward with new topic and areas of IETF is doing we should have visibility of what is going on to make sure it’s not impacted any of the work that the committee is doing. Please give it thought and approach me with who the individual can be.
+
+SHN: We have a strong invited experts list together with AKI and the chairs we’re always reviewing them. We had an extremely busy end of last year. I have had an extremely busy last few months of the end of the year were a bit of a blur if I look back with the activities. I have been very flexible. So the invited experts that typically would have ended their term at the end of 2024, I have not stopped and they will continue on and because we appreciate everybody’s contribution and would like to review them at the end of the year. Sometime in the next little while I will be approaching organizations and members that are related to the organization that are invited experts if they are considered to join. Please be aware to get an email from me. It is a touch base just to inquire and see how your organization is looking at Ecma and if they would like to join because ideally that is the way we want to move forward. So continue on with your work and together with AKI we will work with the chairs to make sure that the list remains accurate.
+
+SHN: With W3C we have done the transition of Winter CG to the generally used term WinterTC and new TC55 that is formed at Ecma. In doing so, I had a number of conversation with W3C leadership and also have met with their CEO Seth Dobbs and they’re keen to see how Ecma can be more engaged with W3C and I know there are already members representative here that are active in different spaces of W3C. Those are the four bullets I brought up in a previous meeting. I would like indication on the **horizontal review** is the one that is the lowest hanging fruit. Does it in any way impact us? I would like guidance. It is important. I want feedback on that. I’m looking at Shane right cross from me. If there are other topics that are important, reach out to me and AKI and the relationship with W3C after the moving of TC55 opens up more opportunities for more collaboration. There may be other projects going on in W3C to come to Ecma. There are common members and eventual invited experts. W3C has a broad scope and a certain focus. Ecma has a value it can bring. If there are topics, we have the opportunity to continue the strong relationship here. We have the liaison contact Michael Smith. I and AKI are active with W3C and have a strong relationship and make sure we are active with W3C and TC55.
+
+DE: Great presentation. A couple comments that I wanted to add. Just to emphasize Winter TC or TC55 the significance of this it’s about standardizing an initially a subset of the web platform that’s supported on web interoperable server environments such as hopefully Node.js and Deno and complementary of TC39 work and scope of things and certain things we thought about in committee that ended up getting web specs for. Hopefully this can help regularize in the ecosystem some of the core concepts. I hope more TC39 members will be interested in joining this group. Development has switched completely to Ecma unlike previous plans where the CG and the TC would live side by side. And thanks LCA and ABO and OMT for the leadership. For the IETF leadership, role, although the IETF have a broad scope and it can be intimidating to be liaison for the whole organization, there’s starter tasks that people could work on. For example, we have certain line types with Ecma that are through registered through IETF such as CycloneDX and small tasks and coordinate to register to the cycle NDX types to point to the new version of the standard. There’s been successful interaction of IETF and TC39 such as making the new version of the datetime format standardized and I hope people consider taking on this extra role.
+
+SHN: Thank you DE. And give applause for the TC55 team. Thank you. I’ll go quickly through the annex slides and then open for any other questions. The annex slides as you see are up loaded. Also reminding everybody of the invited expert and the role on the TC and reminding everybody on the code of conduct which is very important as already noted by RPR on how we work and how we exchange our interaction and work together and collaborate. As always the summary and conclusion are extremely important. It has been an improvement and I thank you very much for all of that and hope to continue to build on that. AKI does the minutes and then I finalize them. So it’s really helpful for us to have these and for everybody else.
+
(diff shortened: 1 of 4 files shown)


meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+* [proposal](https://github.com/tc39/proposal-record-tuple)
+* [slides](https://docs.google.com/presentation/d/1uONn7T91lfZDV4frCsxpwd1QB_pU3P7F6V2j9jEPnA8/)
+
+ACE: Hi, everyone. It’s been a while since we talked about Records and Tuples. Almost a year. I thought it would be a good one first to chat about it again. So it’s like technically Stage 2 proposal, but let’s not worry about that too much. They noticed that whenever a TC39 proposal makes it on to hacker news, it’s almost investible you can put money on someone in the comments will be asking about the status the records and tuples. And also, just on the actual repo itself, people keep asking, what is the status of this. And I am not sure what it is, whether I still like in the design space. So I am going to present some ideas today, I am not necessarily proposing the Stage 3, like NRO told me to do. I am more trying to just kind encourage discussion. Especially from people—from everyone, especially from people have not anything about, I would love to encourage new voices in this area as well. Because we have been talking about this for four or five years now. And new voices are always very, very welcome.
```
acutmore:
```suggestion
ACE: Hi, everyone. It’s been a while since we talked about Records and Tuples. Almost a year. I thought it would be a good one first to chat about it again. So it’s technically a Stage 2 proposal, but let’s not worry about that too much. NRO noticed that whenever a TC39 proposal makes it on to hacker news, it’s almost inevitable that someone in the comments will be asking about the status of records and tuples. And also, just on the actual repo itself, people keep asking "what is the status of this?". And I am not sure what that status is exactly. So I am going to present some ideas today, I am not proposing this for Stage 3 today. I am more trying to just encourage discussion. Especially from people—from every one of course, but I would love to encourage new voices in this area as well. Because we have been talking about this for four or five years now. And new voices are always very, very welcome.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+[Slide: https://docs.google.com/presentation/d/1uONn7T91lfZDV4frCsxpwd1QB_pU3P7F6V2j9jEPnA8/edit#slide=id.g334e668a325_0_0]
+
+ACE: So to catch people up, that haven’t been following kind of all the things over the years, for many years, the proposal was all around adding new primitives. And they had === semantics and typeof that included comparisons here where the special casing things for connecting zero. And then we thought we were kind getting ready for Stage 3, we were kind of changing little bits of the proposal. But these fundamentals things had not changed for a long time. And these fundamentals actually turned out, not appetite to do them. It’s just—for various reasons. Since then we have been a I will bit back to the drawing board on what can we do here? Because I am convinced we can do something. I think this is something that is lacking in the language. And I am sure there is something, that we can do. I just don’t know what it is yet.
```
acutmore:
```suggestion
ACE: So to catch people up, that haven’t been following kind of all the things over the years, for many years, the proposal was all around adding new primitives. And they had === semantics and typeof that included comparisons here where there is special handing for negative zero. And then we thought we were kind getting ready for Stage 3, we were kind of only changing little bits of the proposal. These fundamentals things had not changed for a long time. And these fundamentals actually turned out, there was not appetite to do them. It’s just—for various reasons. Since then we have been back to the drawing board on "what can we do here?". Because I am convinced we can do something. I think this is something that is lacking in the language. And I am sure there is something, that we can do. I just don’t know what it is yet.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+[Slide: https://docs.google.com/presentation/d/1uONn7T91lfZDV4frCsxpwd1QB_pU3P7F6V2j9jEPnA8/edit#slide=id.g2af82517ce6_0_26]
+
+ACE: So a thing I like, so I went off out in the British desert and meditated, if was up to me, like a special birthday present, ACE you get one freeway design and put it to Stage 4.7, what would I choose?
+
+ACE: I would choose that there is syntax. I will comment in more detail. I like the syntax. Maybe it doesn’t have to be this sign. I really think we are running out of ASCII characters to be many other things. Initially, it was a bit of a blow, when we were told we can’t have new primitives, but I am come around to that. Peter hoodie said putting the implementation complexity aside as a JavaScript user, I have come around to that, yes, these things should be objects. I think they should be general containers that contain anything. And I think it’s a great opportunity that they also work as a composite key.
```
acutmore:
```suggestion
ACE: I would choose that there is syntax. I will comment in more detail. I like the syntax. Maybe it doesn’t have to be this sign. I really think we are running out of ASCII characters to be many other things. Initially, it was a bit of a blow, when we were told we can’t have new primitives, but I am come around to that. PHE said putting the implementation complexity aside, as a JavaScript user to would be confusing to have values that look like objects and arrays but are not, I have come around to that, yes, these things should be objects. I think they should be general containers that can contain anything. And I think it’s a great opportunity that they also work as a composite key.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
...ere is syntax was just about freezing objects, which is there is like set proposal for syntax that is just about freezing and sealing. Then I think there are—the advantages for lots of different actors here, I think reading it, at least, I don’t know how people’s brains work. But when I am looking at this, just the view of parentheses, the noise around the data makes it much more readable. It’s fairly less characters to type. But I think it’s beneficial for the reader, there’s more weight on the reader and the writer. And a tooling team, I like this from a tooling perspective. It’s much easier to just analyze this code and see this is a frozen object. It’s not going to change. As opposed to tracking. If `Object.freeze` is being used, tools, I think Rollup, special cases, calls to `Object.freeze`. If there’s enough indirection, that breaks down, creating an `Object.deepFreeze` utility static analysis tools can’t pierce through that to realize that is happening. Syntax gives you the guarantee that no monkey patching can happen. There’s another advantage, which kind of only makes sense later on. I think there is potentially some kind after runtime advantage. But it doesn’t make sense.
```
acutmore:
```suggestion
ACE: So syntax. So yeah. There could be no syntax. To say if there is syntax was just about freezing objects, which is there is like set proposal for syntax that is just about freezing and sealing. Then I think there are—the advantages for lots of different actors here, I think reading it, at least, I don’t know how people’s brains work. But when I am looking at this, just the view of parentheses, the noise around the data makes it much more readable. It’s fairly less characters to type. But I think it’s beneficial for the reader, there’s more weight on the reader and the writer. And a tooling team, I like this from a tooling perspective. It’s much easier to just analyze this code and see this is a frozen object. It’s not going to change. As opposed to tracking. If `Object.freeze` is being used, tools, I think Rollup, special cases, calls to `Object.freeze`. If there’s enough indirection, that breaks down, creating a `deepFreeze` utility static analysis tools can’t pierce through that to realize what is happening. Syntax gives you the guarantee that no monkey patching can happen. There’s another advantage, which kind of only makes sense later on. I think there is potentially some kind after runtime advantage. But that only makes sense once I get to the later slides.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
...the language. You might have some utility that is overloaded. You can maybe pass it. A Number or Array of Numbers and the way it works out the overload is using `Array.isArray`. And now, if we have these tuple things, like arrays but they don’t pass `Array.isArray`, then that utility no longer works. So you would have to update or not use this thing. So if these things are objects and they are arrays and they are inherent from the prototypes that others might expect, there’s a larger chance that you could just adopt these. I think there’s also a benefit to people learning the language, the typeof is something tends to be something that immediately, like, Chapter 1, when you open the book, these are the call parts of the language. More like, think, immutable data structures are really, really, really useful and something you want to learn about sooner than later. I don’t it’s necessarily Chapter 1, page 1 concepts. But overall, I think I will like them being objects. The reason they weren’t objects in the proposal is because it helps explain from a kind of modelling aspect of how the language collects some other behaviors. But I think the weighting is actually crucial in the matrix.
```
acutmore:
```suggestion
ACE: So I also, as I said, I have really come around to these being objects because again the point that PHE made was, if you are going to adopt these things, there’s already a lot of code out there that is kind of sniffing, you’re going to reflect the properties of values in the language. You might have some utility that is overloaded. You can maybe pass it a number or array of numbers and the way it works out the overload is using `Array.isArray`. And now, if we have these tuple things, like arrays but they don’t pass `Array.isArray`, then that utility no longer works. So you would have to update or not use this thing. So if these things are objects and they are arrays and they inherent from the prototypes that others might expect, there’s a larger chance that you could just adopt these. I think there’s also a benefit to people learning the language, the typeof is something tends to be something that immediately, like, Chapter 1, when you open the book, these are the core parts of the language. More like, think, immutable data structures are really, really, really useful and something you want to learn about sooner than later. I don’t it’s necessarily Chapter 1, page 1 concepts. But overall, I think I will like them being objects. The reason they weren’t objects in the proposal is because it helps explain from a kind of modelling aspect of how the language collects some other behaviors. But I think the weighting is actually crucial in the matrix.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
...eat. But it really cuts off a lot of language. Like, maybe for designing the language from scratch, we could do that. But the kind has already sailed. Pretty much everything in the language is mutable unless you do work to stop it being that way. And I just felt really sorry for the people that are going to use the new things we are building, like Temporal, which is fantastic, and has an immutable data model. So if someone thinks I am using immutable data, these aren’t like old school dates, where they are internal mutable and you can change what type they point to. Everything in Temporal, data model is these things. Are mutable, but mutable objects. You can add new properties to them. From like a perspective of records and tuples, the thing has to be frozen or stabled or fixed, like from marks, then it would cutoff large parts of the language, even things from a user’s perspective, do follow the roll of being I am mutable. And I think maybe they’re still ways that, you know, linting tool might be spot little mistakes where people have immutability. We talked about having a box like object that lets you opt out, you have to code into the data model and makes it much harder to adopt
```
acutmore:
```suggestion
ACE: So the other part about the proposal is that it was kind of always about we found the original commits. 2019, and back then, it was—these things are deeply immutable. The only thing you could put in them were other Records and Tuples and primitives. And again the feedback we got on that was that’s great. But it really cuts off a lot of language. Like, maybe for designing the language from scratch, we could do that. But the ship has already sailed. Pretty much everything in the language is mutable unless you do work to stop it being that way. And I just felt really sorry for the people that are going to use the new things we are building, like Temporal, which is fantastic, and has an immutable data model. So if someone thinks "I am using immutable data", these aren’t like old school dates, where they are internal mutable and you can change what type they point to. Everything in Temporal, data model is these things are immutable, but still actually mutable objects. You can add new properties to them. From like a perspective of records and tuples, the thing has to be frozen or stabled or fixed, like from MM, then it would cutoff large parts of the language, even things from a user’s perspective, do follow the roll of being I am mutable. And I think maybe they’re still ways that, you know, linting tool might be spot little mistakes where people have immutability. We talked about having a box like object that lets you opt out, you have to code into the data model and makes it much harder to adopt.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+WH: Okay.
+
+SYG: Yeah. DE mentioned initializer lists. The biggest—I agree. That it’s pretty cool to have immutable structs, shared structs notwithstanding for now. The problem is basically, if you want user constructors, like the way it works today in the proposal is that you get this one shot initialization by the time any of this value is escaped to user code it has the own properties that are declared already on it. And then, you know, you can mutate it. But if you want something to be born immutable, that won’t work. So then you have the problem of, well, how do you limit the access to this thing while it’s in this initialization phase, after that is over, it is then immutable.
+
+ACE: Yeah
+
+SYG: I don’t know the solution to that. If we do, that would be nice.
+
+ACE: Yeah. But a similar thing in JavaScript, you can—for them it’s easier because the syntax isn’t this stock filled. You can say fill = and it understands that. Like, I have ideas in mind. I can share in Matrix later. Please the possibility of an immutable structs is around.
```
acutmore:
```suggestion
ACE: Yeah. But a similar thing in Java's Valhalla, for them it’s easier because the syntax isn't `this.field =`. You can say `field =` and it understands that. Like, I have ideas in mind. I can share in Matrix later. I'm pleased the possibility of an immutable struct is around.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+ACE: The other kind of downside of keys—let’s say records and tuples, similar to what Shu was saying, if these were shared structs there’s a cost to allocating shared structs. If you were creating the immutable data structs, just to get the immutableness of them, and not because they are planning on doing `===`, they will pay a cost creation time that they will never claim. You have to do this work eagerly to do this at all. That’s a bit after shame. People might say “please don’t use this here because they’re too slow to create”.
```
acutmore:
```suggestion
ACE: The other kind of downside of keys—let’s say records and tuples, similar to what SYG was saying, if these were shared structs there’s a cost to allocating shared structs. If you were creating the immutable data structs, just to get the immutableness of them, and not because they are planning on doing `===`, they will pay a cost creation time that they will never claim. You have to do this work eagerly to do this at all. That’s a bit after shame. People might say “please don’t use this here because they’re too slow to create”.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+ACE: I’m pretty sure this causes GC complexity that we could see as an opportunity for some really good GC research or say that’s the reason we wouldn’t go in this direction. And you have to—I think there are people in committee that want negative zero and in turn means you don’t get negative zero. It’s another thing that the MDM library always get wrong. They always forget about negative zero.
```
acutmore:
```suggestion
ACE: I’m pretty sure this causes GC complexity that we could see as an opportunity for some really good GC research or say that’s the reason we wouldn’t go in this direction. And you have to—I think there are people in committee that want negative zero and in turn means you don’t get negative zero. It’s another thing that the NPM library always get wrong. They always forget about negative zero.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+[Slide: https://docs.google.com/presentation/d/1uONn7T91lfZDV4frCsxpwd1QB_pU3P7F6V2j9jEPnA8/edit#slide=id.g336fbd25823_0_47]
+
+ACE: So going back to my meditating in the desert image, I was thinking you don’t necessarily have to intern these thing. What we have is an opportunity. If we added these to the language, we kind of get this one time or this one opportunity to decide what the semantics are. And we could say that these new things added to the language have new behavior, like, they introduce a new capability to steal a word from DE’s slides that haven’t been presented yet, we can say in particular APIs in the language even if the things are triple equal to each other, they could still be treated equal in certain APIs.
```
acutmore:
```suggestion
ACE: So going back to my meditating in the desert image, I was thinking you don’t necessarily have to intern these thing. What we have is an opportunity. If we added these to the language, we kind of get this one time or this one opportunity to decide what the semantics are. And we could say that these new things added to the language have new behavior, like, they introduce a new capability to steal a word from DE’s slides that haven’t been presented yet, we can say in particular APIs in the language even if the things are not triple equal to each other, they could still be treated equal in certain APIs.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+ACE: So maybe we could actually say when you use these things in a map, the map kind of snips them, sees they’re records and tuples and applies the notation and this line here would work.
```
acutmore:
```suggestion
ACE: So maybe we could actually say when you use these things in a map, the map kind of checks them, sees they’re records and tuples and applies the notation and this line here would work.
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+### Speaker's Summary of Key Points
+
+It sounds like most people would strongly prefer that 1) this produces a data property 2) That this does not interfere with or touch the `[[ErrorData]]` internal slot should it exist. I will pursue those choices.
+
+### Conclusion
+
+Stage 1 advanced
+
+## Meta discussion
```
bakkot:
```suggestion
## Discussion about shipping non-standard features
```
maybe? "meta discussion" isn't very informative.

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+
+ACE: It’s not clear to me either. I think they can’t be, because I don’t think we’ll be able to – I’m putting words in implementer’s mouth. I think implementers will say, yes, that will be too slow because maybe there’s a way that in-turning costs could be reduced. From everything I’ve been told and everything I kind of understand about the engines, I feel like the same view that we adopt from these being primitives applies here and that them being triple equal would be nice from a semantic point of view but I think the fact that these things are actually going to run and be performative means they wouldn’t be.
+
+DE: I’ve been assuming they’re not triple equals as well. And this is because the previous time that records and tuples were proposed for Stage 3, we got extremely explicit feedback from implementers that neither strategy would work, not in-turning because the cost is too high and doing the deep comparison in place because it’s too important that triple equals on objects is just a pointer comparison. So the proposal that Ashley is talking about tries to work within that
```
jayaddison:
```suggestion
ACE: It’s not clear to me either. I think they can’t be, because I don’t think we’ll be able to – I’m putting words in implementer’s mouth. I think implementers will say, yes, that will be too slow because maybe there’s a way that interning costs could be reduced. From everything I’ve been told and everything I kind of understand about the engines, I feel like the same view that we adopt from these being primitives applies here and that them being triple equal would be nice from a semantic point of view but I think the fact that these things are actually going to run and be performative means they wouldn’t be.

DE: I’ve been assuming they’re not triple equals as well. And this is because the previous time that records and tuples were proposed for Stage 3, we got extremely explicit feedback from implementers that neither strategy would work, not interning because the cost is too high and doing the deep comparison in place because it’s too important that triple equals on objects is just a pointer comparison. So the proposal that Ashley is talking about tries to work within that
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
...
+| Sergey Rubanov   | SRV          | Invited Expert     |
+| Justin Grant     | JGT          | Invited Expert     |
+| Ron Buckton      | RBN          | Microsoft          |
+| Nicolò Ribaudo   | NRO          | Igalia             |
+| Jesse Alama      | JMN          | Igalia             |
+| Samina Husain    | SHN          | Ecma               |
+| Istvan Sebestyen | IS           | Ecma               |
+| Eemeli Aro       | EAO          | Mozilla            |
+| Aki Rose Braun   | AKI          | Ecma International |
+
```
js-choi, line 34:
```suggestion
| J. S. Choi       | JSC          | Invited Expert      |

```
I was present on Wednesday; apologies for not spotting this omission earlier.

meetings/2025-02/february-20.md
```diff
@@ -0,0 +1,800 @@
...
+
+JSC: Just adding on to what JSL said concerns about actual unanimous blocking makes sense. I’m just talking about having some sort of temperature check showing people, telling them if I were blocking, if I blocked, would someone else also block too and then you have to justify everything and whatever. If you know the answer yourself and predicting what SYG is going to say. If you don’t want to propose to advance, you don’t need this question or whatever. But it still could be useful, I guess, for other people who would want to. But if you already know you don’t want the proposal to advance, then you can just block it or whatever any way. That’s all.
```
js-choi:
```suggestion
JSC: Just adding on to what JSL said. The concerns about actual anonymous blocking make sense. I’m just talking about having some sort of temperature check showing people, telling them if I were blocking, if I blocked, would someone else also block too and then you have to justify everything and whatever. If you know the answer yourself and can predict what the other is going to say…If you don’t want the proposal to advance, you don’t need this question or whatever. But it still could be useful, I guess, for other people who would want to. But if you already know you don’t want the proposal to advance, then you can just block it or whatever anyway. That’s all.
```

meetings/2025-02/february-20.md
```diff
@@ -0,0 +1,800 @@
...
+
+JSC: We found temperature checks using TCQ with emoji from what I recall. Exactly like that. But I think it’s important to have it be shown with a specific question like would you block this? And have it remain completely anonymous. And positive, positive and following, and confused, instead of something like this, I would like it to be clear, would you block this and have it remain completely unanimous? I think that would be productive. Does that go with what you’re saying? And mitigate the reluctance of people to show and give information unanimously to people whether they would be the only one who would block it or not.
```
js-choi:
```suggestion
JSC: We have temperature checks using TCQ with emoji from what I recall. Exactly like that. But I think it’s important to have it be shown with a specific question like would you block this? And have it remain completely anonymous. And positive, positive and following, and confused, instead of something like this, I would like it to be clearly “would you block this” and have it remain completely anonymous. I think that would be productive. Does that go with what you’re saying? And mitigate the reluctance of people to block, by giving information anonymously to people whether they would be the only one who would block it or not.
```

js-choi:
I was reading the notes again today and noticed some errors. I’ve left comments in a review. My sincere apologies for the belatedness.

meetings/2025-02/february-20.md
```diff
@@ -0,0 +1,800 @@
...
+|------------------|--------------|---------------------|
+| Chris de Almeida | CDA          | IBM                 |
+| Samina Husain    | SHN          | Ecma                |
+| Eemeli Aro       | EAO          | Mozilla             |
+| Daniel Ehrenberg | DE           | Bloomberg           |
+| Daniel Minor     | DLM          | Mozilla             |
+| Ujjwal Sharma    | USA          | Igalia              |
+| Art Vandelay     | AVY          | Vandelay Industries |
+| Jesse Alama      | JMN          | Igalia              |
+| Ron Buckton      | RBN          | Microsoft           |
```
js-choi, line 15:
```suggestion
| Ron Buckton      | RBN          | Microsoft           |
| Nicolò Ribaudo   | NRO          | Igalia              |
| Kevin Gibbons    | KG           | F5                 |
```
Nicolò was present on Thursday, as can be seen from the presence of “NRO” in this day’s logs.
Kevin Gibbons was present on every day.

meetings/2025-02/february-18.md
```diff
@@ -0,0 +1,1553 @@
+# 106th TC39 Meeting | 18 February 2025
+
+**Attendees:**
+
+| Name             | Abbreviation | Organization       |
+|------------------|--------------|--------------------|
```
js-choi, line 6:
```suggestion
|------------------|--------------|--------------------|
| Kevin Gibbons    | KG           | F5                 |
```

meetings/2025-02/february-19.md
```diff
@@ -0,0 +1,1327 @@
+# 106th TC39 Meeting | 19 February 2025
+
+**Attendees:**
+
+| Name             | Abbreviation | Organization       |
+|------------------|--------------|--------------------|
```
js-choi, line 6:
```suggestion
|------------------|--------------|--------------------|
| Kevin Gibbons    | KG           | F5                 |
```

michaelficarra:
@js-choi I don't think you published the review.

js-choi (changes requested):
Sorry, forgot to actually submit the review.

ctcpip (approved):

PR #365, opened 2025-03-23
Update delegates.txt

brad4d:
Add abbreviation for Stephen Hicks / Google


ctcpip:
@shicks would you be ok with using `SHS` to match our convention (first letter of first name, first and last letter of last name)?

shicks:
Yes, that's fine.

ctcpip (approved):

PR #372, opened 2025-05-28
add Zbigniew Tenerowicz to delegates.txt

naugtur:
Adding self with an abbreviation. ZB was already taken


nicolo-ribaudo:
Hi! Normally the abbreviations are "first letter of the name", "first letter of the surname", "last letter of the surname". If you don't have a super strong preference I recommend going with ZTZ :)

naugtur:
I react to ZB so ZBT was a way to call my attention and ZTZ is not, but I'd rather follow a convention.

ryzokuken (approved):
thanks, we can prompt you with "Zibi" while your TLA in the notes would be ZTZ in this case

PR #375, opened 2025-06-24
Add Elliot Goodrich to delegates.txt

elliotgoodrich:


evenstensberg (approved):

jkup:
New delegate checklist is waiting for review right now

linusg (commented):
I can't find the document that describes it right now it but the correct acronym would be EGH.

linusg:
Here it is: https://github.com/tc39/how-we-work/blob/main/how-to-take-notes.md#the-attendee-list

> Normally we use three letters: first letter of first name, first letter of last name, last letter of last name. Old-timers are grandfathered-in with two-letter initials.

elliotgoodrich:
Thank you @linusg - I have updated the initialism.

michaelficarra (approved):

PR #378, opened 2025-07-17
chore: update delegates.txt

styfle:
As requested from Michael in Matrix


ljharb:
Our typical convention would yield "SST" here

styfle:
I prefer `STY` since its closer to @styfle whereas `SST` doesn't have any identity

styfle:
I would also be open to `SS` but I'm not sure if two letters are still allowed.

ljharb (approved):
I think two letters are off the table, but I'm fine with "STY".

PR #380, opened 2025-09-08
Add Satish Srinivasan to delegates list.

aardvark179:


delegates.txt
```diff
@@ -437,6 +437,7 @@ Sarah D'Onofrio (SDO)
 Sarah Groff Hennigh-Palermo (SHO)
 Sathya Gunasekasan (SGN)
 Satish Chondra (SC)
+Satish Srinivasan (SRI)
```
ljharb: our typical convention would have your acronym be "SSN". is that ok @0xe?

0xe: Yes, SRI is okay. Happy with SSN as well.

ljharb (approved):

Issue #245, opened 2023-04-09
repeated Speaker's Summary


jmdyck:
In meetings/2023-03/mar-22.md, for the item "Type Annotations Delimiting Concerns", the "Speaker's Summary of Key Points" is the same as that for the previous item "Type Annotations Proposal Update".

(The 'Conclusion' is also the same, though that's perhaps appropriate.)

aaronliu0130:
@waldemarhorwat was the presenter for https://github.com/tc39/notes/blob/main/meetings/2023-03/mar-22.md#type-annotations-delimiting-concerns

dav3yblaz3:
if the conclusion is the same can we not press forward for progress as a team and follow the suggestion that was offered as we are a global company

PR #416, opened 2026-07-11
docs: fix duplicate speaker summary for type annotations delimiting concerns

MatrixNeoKozak:
Resolves issue #245 by updating the 'Speaker's Summary of Key Points' under 'Type Annotations Delimiting Concerns' in meetings/2023-03/mar-22.md to accurately summarize the discussion on syntax, token soups, and grammar conflicts, rather than duplicating the type erasure summary from the previous section.

### DIFF
--- a/meetings/2023-03/mar-22.md
+++ b/meetings/2023-03/mar-22.md
@@ -1162,10 +1162,10 @@ JRL: I think beyond that there was strong pushback from browsers that runtime ty
 
 ### Speaker's Summary of Key Points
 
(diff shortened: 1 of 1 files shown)


ptomato:
This PR seems to miss the point of the issue - the two conclusions sections are not supposed to be the same, as they are conclusions of two different discussions. With this PR they are still duplicated, only re-summarized. Please also note the [AI policy](https://github.com/tc39/how-we-work/blob/main/AI_POLICY.md) which applies to prose contributions.

ctcpip:
It's called the "Speaker's Summary". So unless you are the Speaker, you should not make updates to it.

